PR #237, opened 2019-03-27
Improve how we update the events

ericpoirier:


chrisguindon:
As discuss, I am hoping that this is something that we can implement in our base theme since I don't consider this feature specific to jakarta.ee:
https://github.com/EclipseFdn/hugo-solstice-theme

Issue #255, opened 2019-04-10
Jakarta EE Compatible Product page


TanjaObradovic:
**Is your feature request related to a problem? Please describe.**
Can we please create 
 jakarta.ee/compatibility
a page where we would list all products that are Jakarta EE certified? 

**Describe the solution you'd like**
Once we have the page we want to list all Compatible Implementations of Jakarta EE.
For Compatible Software Products to be listed on the Jakarta EE official list of Software Products must provide, as a minimum:
- A product name and version,
- A product logo (optional)
- The Participant, Enterprise or Strategic Member name supporting the Compatible Software Product,
- The URL to the published public TCK test results demonstrating compatibility as defined in the TCK Process,
- The URL where Compatible Software Product is available for download

A request to list a product can be made by creating a GitHub issue that will request the above information.

ericpoirier:
Hi Tanja,

Do you have some content that we can add to this page. The reason I'm asking, is so we don't create an empty page on the website.

Thanks.

TanjaObradovic:
Can we just put some text, description on what this page is going to contain?
Something like: "This page is to list Jakarta EE downloadable compatible implementations once Jakarta EE 8 release becomes available". If you think we need more then this, let's involve Thabang and Stephanie.

Thanks!

TanjaObradovic:
Hi Eric,
can we put something like this on the page:

This page is to list downloadable Compatible Implementations of Jakarta EE. For a Compatible Implementation to be listed on the Jakarta EE official list please follow this two-step process:
1. Send an email to tck@eclipse.org with a pointer to the GitHub issue described in the next step
2. Open a Git Hub issue
For Compatible Software Products to be listed on the Jakarta EE official list of Software Products must provide, as a minimum:
A product name and version,
A product logo (optional)
The Participant, Enterprise or Strategic Member name supporting the Compatible Software Product,
The URL to the published public TCK test results demonstrating compatibility as defined in the TCK Process,
The URL where Compatible Software Product is available for download

Thanks,
Tanja

ericpoirier:
Hi Tanja,

Would you like this new page to have a link in the main menu or only in the sidebar of subpages?

Thanks

TanjaObradovic:
We could have an item next to "blog" column up top  (About, Membership, Connect, Resources, Blog) that will lead to the page on its own. I will ping Stephanie to review and let us know if she disagrees.

Thanks!

ericpoirier:
[...]

PR #260, opened 2019-04-22
#255 - Add compatibility page and issue template

ericpoirier:

### DIFF
--- a/.github/ISSUE_TEMPLATE/compatibility.md
+++ b/.github/ISSUE_TEMPLATE/compatibility.md
@@ -1,0 +1,17 @@
+---
+name: Compatible Software Products
+about: For Compatible Software Products to be listed on the Jakarta EE official list of Software Products
+
+---
+
+For Compatible Software Products to be listed on the Jakarta EE official list of Software Products must provide, as a minimum:
+
+**A product name and version**
+
+**A product logo (optional)**
+
+**The Participant, Enterprise or Strategic Member name supporting the Compatible Software Product**
+
+**The URL to the published public TCK test results demonstrating compatibility as defined in the TCK Process**
+
+**The URL where Compatible Software Product is available for download**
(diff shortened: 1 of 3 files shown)


chrisguindon:
My concern here is that we are adding yet another top level item for a new page.

May I propose that we move "Blog" under "Resources" if we are going to add a "Compatibility" menu item?

@stephanieswart @TanjaObradovic What do you think?

TanjaObradovic:
I will let Stephanie have the final say on this one. I think I am ok with it, but it is nice to have "blog" on it own :). I guess Jakarta EE is growing and it makes sense to move things around.

stephanieswart:
Personally, I don't think we should hide blogs under resources as I don't think that a blog is a resource, it's an ongoing center of information. I think there will be less visibility on it unless maybe the community news section could automatically update with every new blog post that comes in. However, if we'd like to remove blog from the navigation bar, I think we need to implement something to make up for the lack of visibility.

TanjaObradovic:
+1 with Stephanie's comment

chrisguindon:
The problem that I am trying to highlight is that we can't add a new top level item for each new page since will run out of space. 

I think this issue is showing us that we need to improve our content strategy.

For example, eclipse.org is a website with way more content and we only have 3 top level links (Members, Working Groups and Projects). 

Perhaps it's time to bring back the more menu to include links such as all the ressources. I personally would move blogs under the more menu as we are doing with www.eclipse.org and planeteclipse.org.

Also,  the "Specification Process" section in the resource dropdown is not very useful. We are linking to a google drive documents that is NOT public and to a CFP proposal from 2018.

chrisguindon:
Let's merge this change since @TanjaObradovic is waiting for this to go live but I am hoping that the marketing team will take in considerations my recommendations for the future!

Issue #302, opened 2019-06-07
Define the means by which we promote Final Specifications


waynebeaton:
We need to define a means by which specification content is delivered. As core content, Final Specifications (as defined by the Eclipse Foundation Specification Process) must must be hosted on Eclipse Foundation supported infrastructure (with qualities of services that include, for example, backups).

With the successful conclusion of a release review for a specification project, the project team will deliver a Final Specification package that includes a specification document in PDF and HTML; and a collection of technical artifacts that may include API JARs, a TCK, and Javadoc.

Here's what I propose that we do:

To make that content available to consumers, I propose that we configure a directory structure on the web server along these lines: specs/[short-name]/[version]/*. Note that this is per-specification, not per-project. 

e.g. specs/faces/2.3/* for Jakarta Server Faces 2.3, and specs/faces/2.4/* for Jakarta Server Faces 2.4.

The specification document, and technical artifacts will just be files in the directory; the Javadoc (HTML files) will be expanded out in a directory so that they can be browsed and index as normal. No access restrictions are required for any of this content. 

I believe that it would be best that we not check these into a Git repository to stage them: accepting a pull request against this content would be fundamentally wrong. The content is not going to be maintained here, this is exclusively a distribution channel. Changes to this content must be managed via the Eclipse Foundation Specification Process. Also note that some of this content will be huge (I believe that the TCK for Jakarta EE Platform is something like 800GB).

For each specification, we will create a landing page. That landing page will be a Hugo file in a directory in this repository. That file will initially contain information about only the Jakarta EE 8 release of the specification, but will be expanded to include all future release versions of the specification (Final Specification).

For each version, we'll capture pointers to the aforementioned content, along with other related metadata, for example:

* Specification name and version;
* Name and URL for specification license (Eclipse Foundation Specification License);
* Project name and URL to project website;
* Contact/communication channel links;
* Pointer to normative document (the PDF); pointer to the HTML (labeled as "non-normative");
* Pointer to the TCK and TCK instructions content;
* Name and URL for TCK license (Eclipse Foundation TCK License);
* List of Compatible Implementations (which can be updated);
* ...

For each version, we will also need to provide a public record of results from various votes related to the project.

At some point, we should consider creating a custom Hugo content type. This will allow us to specify this metadata in a manner that is machine readable. This will make it easier to build aggregation pages, and manipulate the data in other ways (e.g. provide a JSON feed of specification information). 

[...]

PR #357, opened 2019-07-17
Define the means by which we promote Final Specifications #302

chrisguindon:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /static/js/
 /static/css/
 /public/
+/content/specifications
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "watch": "NODE_ENV=development webpack --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
     "hot": "NODE_ENV=development webpack-dev-server --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
     "production": "NODE_ENV=production webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "install": "NODE_ENV=production npm run production"
+    "install": "NODE_ENV=production npm run production",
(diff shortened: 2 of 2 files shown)


ggam:
The only problem I see with this approach is that you need to manually
rebuild this repository to get the changes.

chrisguindon:
> The only problem I see with this approach is that you need to manually rebuild this repository to get the changes. 

Correct! This is something that I brought up with @waynebeaton before working on this.

I will investigate if there is away to trigger a build on jakarta.ee each time a commit is done on the specifications repo.

An alternative, is to configure a cron job in travis-ci to rebuild the site every day. It will be important to run some tests on the specification repo to make sure that the content can be successfully compiled by hugo.

chrisguindon:
It seems possible to programmatically trigger a job:

- https://docs.travis-ci.com/user/triggering-builds/
- https://github.com/plume-lib/trigger-travis/

waynebeaton:
What is our status? Can we roll this out?

When we make updates to the /specifications repository, how do we roll out the change?

chrisguindon:
@waynebeaton We can merge it at anytime.

As previously discussed, the specification section will be updated each time the site is rebuilt. From what I understood, this was ok.

If we want to rebuilt the site each time a change is submitted to the specification repo, we should consider my proposed solution https://github.com/jakartaee/jakarta.ee/pull/357#issuecomment-512410368

This is something that I believe we must implement on the specification repo, not this website.

waynebeaton:
> As previously discussed, the specification section will be updated each time the site is rebuilt. From what I understood, this was ok.

It will be the Jakarta EE Specification Committee that will decide when they need the content to be updated (i.e., when the site will need to be rebuilt). By what means should that request be made? e.g., an email or GitHub issue? We'll likely only need it rebuilt a few times over the next six weeks or so.

chrisguindon:
@waynebeaton Github issue would be the way to go for now!

Issue #447, opened 2019-09-09
Update the footer & header featured stories


stephanieswart:
title: Jakarta EE 8 is here!
content: Check out all the info and updates about the Jakarta EE 8 release today!
link: jakarta.ee/release
bgImgSrc: ![jakartaee8_background](https://user-images.githubusercontent.com/34580334/64557123-e666a400-d30e-11e9-9c7c-5d61ec6024d9.png)
startDate: 2019-09-10
endDate: 2019-10-07
type: header

header mockup: ![1](https://user-images.githubusercontent.com/34580334/64556896-5cb6d680-d30e-11e9-8287-a777b815bfa3.png)

title: Jakarta EE 8 is here!
content: Check out all the info and updates about the Jakarta EE 8 release today!
link: jakarta.ee/release
bgImgSrc: ![jakartaee8_background](https://user-images.githubusercontent.com/34580334/64557123-e666a400-d30e-11e9-9c7c-5d61ec6024d9.png)
startDate: 2019-09-10
endDate: 2019-10-07
type: footer

footer mockup:
![2](https://user-images.githubusercontent.com/34580334/64557085-d0f17a00-d30e-11e9-8cba-efbe05d369c3.png)

autumnfound:
@stephanieswart while the footer looks alright, the header looks...weird next to jumbotron imho. WDYT?
![image](https://user-images.githubusercontent.com/23018241/64561334-7eb55680-d318-11e9-8a5b-1dfb760b8ed5.png)

autumnfound:
Preview is here: https://deploy-preview-451--jakartaee.netlify.com/
Just refresh until you get the right stories :)

stephanieswart:
I agree it doesn't look good. @autumnfound could you please use this background instead and use blue font?
![jakartaee8_background (1)](https://user-images.githubusercontent.com/34580334/64563790-3305ab80-d31e-11e9-9ffe-bd29db232749.png)

autumnfound:
Preview has been updated, looks better than before

stephanieswart:
Can we make "Sign up to our Newsletter" white?

autumnfound:
Updated!

chrisguindon:
> I agree it doesn't look good. @autumnfound could you please use this background instead and use blue font?

I think the black font looks better and it's easier to read:
https://deploy-preview-451--jakartaee.netlify.com/

Can we revert that font color change?

chrisguindon:
[...]

PR #451, opened 2019-09-09
Update the footer & header featured stories #447

autumnfound:
Change-Id: I72208116778174a8b4fa6a56fdc17b188f3a05a4


chrisguindon:
@autumnfound why are the heading blue? Was that intended or a bug?

PR #598, opened 2020-01-21
Add TCK Process

ivargrimstad:
Signed-off-by: Ivar Grimstad <ivar.grimstad@eclipse-foundation.org>


chrisguindon:
@ivargrimstad The image is not loading but I will merge your change and submit a PR that will fix the image.

Issue #620, opened 2020-02-04
npm install fails on windows


chrisguindon:
I spent some time training @TanjaObradovic yesterday on how to update the jakarta.ee website.

It was an interesting experience since she was on windows. Our instructions mostly work on windows but the npm install script does fails when it tries to run our build.sh script: https://github.com/jakartaee/jakarta.ee/blob/src/build.sh

We need to re-write this script in node.js so that it windows users can contribute to the site.

chrisguindon:
As a temporary fix, I will remove the specification script from the postinstall hook.

This should allow windows users to install all required dependencies to run the site.

jgallimore:
Issues @thharrison and I ran into here:

* `build.sh` is a *nix script - we probably need a `build.bat` for Windows
* `npm install` is running `NODE_ENV=production npm run production` - the assignment to the NODE_ENV environment variable doesn't work on Windows. A `set NODE_ENV=production` and then `npm run production` get a bit further.
* `npm run production` runs `NODE_ENV=production webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js` - again this seems to fall foul of setting the env variable and running the command all in one. Doing a `set NODE_ENV=production` and then `webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js` seemed to work.

Do we need the environment variable, and if so, is it reasonable for the user to set it themselves, rather than the node scripts doing it for us?

Some brief research suggests the cross-env might be able to solve this for us, and not be a particularly invasive change: https://www.npmjs.com/package/cross-env

chrisguindon:
@flora8984461 Sorry for adding you on this issue. I forgot that @Thharrison  and @jgallimore are working on this issue!

chrisguindon:
@jgallimore,

@flora8984461 has created a PR to add some information about windows in the README.md of our basetheme: https://github.com/EclipseFdn/hugo-solstice-theme/pull/160

jgallimore:
@flora8984461 I'm ok if you want to take this over - I don't want to be a blocker. If you need help, please feel free to give a shout.

PR #626, opened 2020-02-10
npm install fails on windows #620

chrisguindon:
Temporary fix until we have time to fix the shell script or move
this content.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ matrix:
       env: HUGO_ENV=production
 install:
 - npm install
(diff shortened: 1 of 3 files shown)


chrisguindon:
> Should we add a line to the netlify config to run this as well? It would increase the build time to what it has been historically, but omitting it would mean that spec changes wouldn't be visible on netlify.

It's already there!

    "build_netlify": "NODE_ENV=production npm run disallow_robots_txt && npm run production && npm run specifications && npm run fetch_project",

Issue #726, opened 2020-04-14
Tabular view of the Jakarta EE committees meeting minutes per year


TanjaObradovic:
Can we add tabular view of the meeting minutes per year? So we open with tab on 2020 / tab 2019 and then tab 2018.
Please refer to the below page for example
https://jakarta.ee/committees/steering/

PR #742, opened 2020-04-22
Tabular view of the Jakarta EE committees meeting minutes per year #726

autumnfound:

### DIFF
--- a/data/meeting_minutes.yml
+++ b/data/meeting_minutes.yml
@@ -1,803 +1,1034 @@
+years:
+  - 2020
+  - 2019
+  - 2018
 items:
    marketing_committee:
        -
           url: marketing_committee/minutes-marketing-april-30-2020.pdf
           title: April 30, 2020 (pdf)
+          year: 2020
        -
           url: marketing_committee/minutes-marketing-april-16-2020.pdf
           title: April 16, 2020 (pdf)
+          year: 2020
        -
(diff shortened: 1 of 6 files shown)


autumnfound:
Pulled to draft as it was pointed out that this isn't in tabs, d'oh!

autumnfound:
Making changes to the general meeting minutes tab would likely require some design work. Right now, we are separating the meeting minutes on the committee pages by year tabs. On the general page, it is done by committee. 

In terms of design, tabs in tabs aren't necessarily ideal imo, but could likely be done. Having a situation like a nested table header would not be suitable long-term as it would be incredibly busy and could very easily spill into extra rows creating a messy looking interface. One way to get around this is to break out the tabs into multiple tabbed sections with a heading+anchor so that the old links would still bring users to the correct section.

Additionally this would have to be done in the base theme, as that's where this functionality lives.

chrisguindon:
@autumnfound I would like to see the meeting minutes in the .tab-content div organized by year.

chrisguindon:
I agree that tabs are not ideal! 

We should keep them on https://deploy-preview-742--jakartaee.netlify.app/meeting_minutes/ for each committee but we could drop them here: https://deploy-preview-742--jakartaee.netlify.app/committees/steering/ in favor to what I am describing in my previous comment?

PR #754, opened 2020-05-08
Shrink the size of the adopter logos

chrisguindon:
Change-Id: I6762fa309fa501cf893f53ce17628cd3bd20067a

### DIFF
--- a/layouts/shortcodes/jug-adopters-list.html
+++ b/layouts/shortcodes/jug-adopters-list.html
@@ -6,14 +6,14 @@
 {{ end }}
 
 {{ if gt (len ($scratch.Get "adopters")) 0 }}
-<h2 id="adopters">Adopters</h2>
+<h2 id="adopters" class="margin-bottom-20">Adopters</h2>
 <ul class="list-inline">
-	{{ range $scratch.GetSortedMapValues "adopters" }}
-		<li>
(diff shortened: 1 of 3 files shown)


layouts/shortcodes/jug-adopters-list.html
```diff
@@ -6,14 +6,14 @@
...
-		<li>
-			<a href="{{ .homepage_url }}" title="{{ .name }} homepage" target="_blank" >
-				<img class="responsive-img img" src="/images/spec_adopters/{{ .logo }}" alt="{{ .name }} logo" width="100" />
-			</a>
-		</li>
-	{{ end }}
+  {{ range $scratch.GetSortedMapValues "adopters" }}
+    <li>
+      <a href="{{ .homepage_url }}" title="{{ .name }} homepage" target="_blank" >
+        <img class="responsive-img img" src="/images/spec_adopters/{{ .logo }}" alt="{{ .name }} logo" style="padding:10px; max-width:100px;" />
```
autumnfound, line 14: If we're using this in multiple places, why not create a class that applies this style? If we ever needed to modify this or give this treatment to another image, we'd either have to put more styles, or clean up all of these values for a class anyways

chrisguindon: Good point! Done!

PR #791, opened 2020-06-03
Fix #572 - Improve the layout of the specifications pages

### LINKED ISSUE #572: improve formatting of /specifications page
We have a lot of specifications on the /specifications page.
Can we "tighten up" the formatting of that page so that I
don't need to do quite as much scrolling to see all the specifications?

We could consider moving the scope statements to a popup,
or move them to the top of each specification page, leaving
just the specification name on the top level "index" page.
Or maybe just getting rid of some of the whitespace and using
a smaller font for the scope statement would be enough?

Eventually we might want to mark specifications as "deprecated"
or "under development" or "experimental" or other similar attributes,
so we might want to think about how to include that information on
this page.

ericpoirier:

### DIFF
--- a/layouts/specifications/list.html
+++ b/layouts/specifications/list.html
@@ -26,30 +26,48 @@
     </h1>
   {{ end }}
   {{ .Content }}
-  <div class="news-list"> 
-    {{ range .Page.Sections.ByWeight}}
-      <a href="{{- with .Params.redirect_url | default .RelPermalink }} {{.}} {{- end }}" class="media media-link{{ if eq .Section "news" }} match-height-item{{ end }}">
-        <h4 class="media-heading">{{ .Title }}</h4>
-        <p class="media-text">{{ .Params.summary }}</p>
-      </a>
-    {{ end }}
-
-    {{ $section := .CurrentSection }}
-    {{ with .Site.Data.spec_adopters.adopters.projects }}
-      {{ range sort . "name" "asc"}}
-        {{ $project := . }}
-        {{ with $section.File.Dir }}
-          {{ if eq . $project.directory }}
-           <h3>JUG Spec Adopters</h3>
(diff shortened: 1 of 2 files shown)


less/styles.less
```diff
@@ -214,3 +214,10 @@ ul.tri-col {
   -moz-column-count: 3;
   column-count: 3;
 }
+
+.news-list {
```
chrisguindon, line 218: You will want to wrap this with another class since it's changing how we display news on the homepage:
https://deploy-preview-791--jakartaee.netlify.app/

layouts/specifications/list.html
```diff
@@ -26,30 +26,46 @@ <h1>
...
-              {{ range sort $project.adopters  "name" "asc"}}
-                <li><a href="{{ .homepage_url }}" target="_blank"><img src="/images/spec_adopters/{{ .logo }}" alt="{{ .name }}" class="adopter-logo"/></a></li>
-              {{end}}
-            </ul>
+  {{ if eq .Page.Parent.Title "Specifications" }}
+    <ul>
+      {{ range .Page.Sections.ByWeight}}
+        <li>
+          <h3><a href="{{- with .Params.redirect_url | default .RelPermalink }} {{.}} {{- end }}">{{ .Title }}</a></h3>
+          <p>{{ .Params.summary | truncate 57 }}</p>
```
ivargrimstad, line 34: The attribute `60CharSummary` should be used if present. The truncate to 57 is fallback for missing `60CharSummary`. See Jakarta MVC for an example where it is defined.

ivargrimstad: Yes, the parameter can be changed. I will update the MVC spec right away. I don't think any other specs have added this parameter yet.

ivargrimstad: I will use `summary60Char` btw...

PR #790, opened 2020-06-02
Fix #727 - Update class for Duke's Choice Award in featured story block

ericpoirier:
Signed-off-by: Eric Poirier <eric.poirier@eclipse-foundation.org>

PR #794, opened 2020-06-05
update package.json

ericpoirier:

### DIFF
--- a/data/featuredstory.yml
+++ b/data/featuredstory.yml
@@ -1,5 +1,5 @@
 defaultRight: |
-        <div class="featured-ad-content-jakarta">
+        <div class="featured-side-content-jakarta">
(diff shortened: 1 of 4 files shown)


autumnfound:
Side content text is not accessible with new/updated side content background colour. We should either add a var to set the side background colour back to what it was originally, or make the text black to be legible.

chrisguindon:
@ericpoirier How about you close https://github.com/jakartaee/jakarta.ee/pull/790 and bring those changes here.

Make sure to generate a fresh package-lock.json file.

ericpoirier:
> Side content text is not accessible with new/updated side content background colour. We should either add a var to set the side background colour back to what it was originally, or make the text black to be legible.

Not sure what backgroud colour you refer to here. Can you tell me which page you would like to change the background?

ericpoirier:
>     * lock file is missing
Done
>     * the Duke "ad" text is now white and the background is lighter than it should!
 I added the changes to the Duke's award PR here in this PR.

PR #812, opened 2020-06-21
Update batch XSDs to version 2.0

scottkurz:

Following Jakarta EE 9 convention , have the "version" attribute for the top-level element ("job") match the spec version and file name... so bump up from 1.0 to 2.0.

The earlier PR #707 converted the target NS of the schema elements but it left the version as 1.0.


scottkurz:
Closing for now ...  I want to first discuss within batch community whether we should really used 'fixed' or if this hurts support for older Java EE 7 job XML

PR #824, opened 2020-06-23
update Steering and Spec meeting min

TanjaObradovic:
Signed-off-by: Tanja Obradovic <tanja.obradovic@eclipse-foundation.org>


data/meeting_minutes.yml
```diff
@@ -279,6 +279,14 @@ items:
        
        
    specification_committee:
+       -
```
autumnfound, line 282: These 2 entries are reversed, and appear reversed on the site. These aren't currently sorted outside of by year.

TanjaObradovic: Many thx for your help Martin!!!!

chrisguindon:
Good catch @autumnfound

PR #840, opened 2020-07-19
Update batch XSDs to version 2.0

scottkurz:
Signed-off-by: Scott Kurz <skurz@us.ibm.com>

Following Jakarta EE 9 convention , have the "version" attribute for the top-level element ("job") match the spec version and file name... so bump up from 1.0 to 2.0.

The earlier PR #707 converted the target NS of the schema elements but it left the version as 1.0.

PR #841, opened 2020-07-21
Update batch XSDs to version 2.0

scottkurz:

Following Jakarta EE 9 convention , have the "version" attribute for the top-level element ("job") match the spec version and file name... so bump up from 1.0 to 2.0.

The earlier PR #707 converted the target NS of the schema elements but it left the version as 1.0.

The matching PR in 'batch-api': eclipse-ee4j/batch-api#103 has been reviewed (unfortunately we have the files duplicated across the two projects, but changes should be rare).

One slightly less-elegant component of this fix is to define our own version type, as mentioned in the comment.  This avoids including the common Jakarta schema file which doesn't seem worth it for such a simple type definition.   Tried to avoid any namespace collision with the 'batchVersionType' name.

---

(We can ignore the similar PR https://github.com/jakartaee/jakarta.ee/pull/840, which got messed up by me using the wrong base branch.)

### DIFF
--- a/static/xml/ns/jakartaee/batchXML_2_0.xsd
+++ b/static/xml/ns/jakartaee/batchXML_2_0.xsd
@@ -20,7 +20,7 @@
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
     elementFormDefault="qualified"
     targetNamespace="https://jakarta.ee/xml/ns/jakartaee"
-    xmlns:jbatch="https://jakarta.ee/xml/ns/jakartaee" version="1.0">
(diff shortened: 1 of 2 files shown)


scottkurz:
@ivargrimstad ^^^

ivargrimstad:
@chrisguindon The Travis build is successful but doesn't report status back for some reason

https://travis-ci.org/github/jakartaee/jakarta.ee/builds/710340382

chrisguindon:
@ivargrimstad This occurs every now and then. I restarted the build to see if that will fix the issue.

chrisguindon:
@ivargrimstad travis-ci is now reporting that the build was successful!

ivargrimstad:
@chrisguindon  thanks :)

PR #850, opened 2020-07-28
Add Jakarta Contexts and Dependency Injection 3.0 beans.xml schema file

starksm64:
Signed-off-by: Scott M Stark <starksm64@gmail.com>


kwsutter:
This looks good, but the schema page is extremely hard to find... Here's the url for this particular PR:
https://deploy-preview-850--jakartaee.netlify.app/xml/ns/jakartaee/#9 

But, in general, how are users supposed to find the schema page from https://jakarta.ee ?  I tried several different paths with no luck.  I eventually used the Search function for "schemas" and found the url.  But, don't we need a more natural navigation to this page?  @ivargrimstad -- something to look at when you get back from vacation?  Thanks.

starksm64:
> But, in general, how are users supposed to find the schema page from https://jakarta.ee ? I tried several different paths with no luck. I eventually used the Search function for "schemas" and found the url.

We should probably have a Schemas item on the https://jakarta.ee/specifications/ page, and a link in the Resources menu.
It might be good if the platform page for each release (https://jakarta.ee/specifications/platform/9/) also had a link in a Schemas section.

We could also consider adding it to each specification index.md template to be filled in if the schema has templates. I have added one to the CDI index.md page.

ivargrimstad (approved):

PR #874, opened 2020-08-31
Shabs/src/mkt cmte meeting mins 20 august

shabnammayel:


data/meeting_minutes.yml
```diff
@@ -4,6 +4,14 @@ years:
   - 2018
 items:
    marketing_committee:
+        -
```
autumnfound, line 7: There is one too many spaces before the dash on this line, which is likely causing the breakage. The way this file is written is super picky with the consistency of spacing

PR #888, opened 2020-09-15
add ZH config && pages for jakarta.ee

laliwa:
Signed-off-by: laliwa <40311795@qq.com>


layouts/shortcodes/pages/compatibility-products.zh.html
```diff
@@ -0,0 +1,48 @@
+<div class="row">
```
autumnfound, line 1: Rather than create a file for this, we should add new i18n values for the content that needs to be multilingual, and update the base shortcode instead. The new i18n values can be added using a new i18n folder at the root level, and 2 files added to that folder, en.toml and zh.toml. These files will combine with the files we merged into the base theme a week or two ago! More info on how this all works can be found on the [Hugo tutorial/help page for internationalization](https://gohugo.io/content-management/multilingual/#translation-of-strings).

The same page mentioned above will describe the usage of those values as well! We would need something like `{{ i18n "compatability.intro.text" }}`, `{{ i18n "compatability.logo.alt" }}`, and `{{ i18n "compatability.proof-of-compatibility.text" }}`.

This file in particular may need more attention around vendor/product names. I'm not 100% sure how we would handle most of this, as it's IP stuff that's outside of my normal realm.

layouts/shortcodes/pages/connect.zh.html
```diff
@@ -0,0 +1,123 @@
+<h2 class="text-center heading-line"><span>邮件列表</span></h2>
```
autumnfound, line 1: See above comment. If we use i18n files, the management of shortcodes, and changes made to them become way easier!

autumnfound (changes requested):
In general, we should make a few fixes around shortcodes and i18n translations. I can look at this more tomorrow and submit a patch to this PR to help out with this!

layouts/shortcodes/pages/jakartaeechina/index.zh.html
```diff
@@ -0,0 +1,34 @@
...
+<h3 id="out">成员列表</h3>
+<p class="memberlist" style="line-height:2;"></p>
+
+<script src="https://apps.bdimg.com/libs/jquery/1.6.4/jquery.min.js" type="text/javascript"></script>
+
+<script>
+$.ajax({ 
+		type:"get",
+        dataType:"json",
+		url: "http://jakarta.eaworld.io/belisted/getlist.php",
```
chrisguindon, line 25: Netlify detected Insecure mixed content. This needs to change to https.

layouts/shortcodes/pages/jakartaeechina/index.zh.html
```diff
@@ -0,0 +1,34 @@
...
+<p class="margin-top-20"><a class="btn btn-lg btn-secondary" href="http://jakarta.eaworld.io/belisted/mailinglist.php">Jakarta EE 中国社区邮件列表</a></p>
+
+<h3 id="out">想要把你的名字也加入列表么？</h3>
+<p>目前Jakarta EE中国社区已经正式成立，我们欢迎所有对Java EE或Jakarta EE感兴趣的同学加入社区。</p> 
+<p class="margin-top-20"><a class="btn btn-lg btn-secondary" href="http://jakarta.eaworld.io/belisted">加入 Jakarta EE 中国社区</a></p>
+
+<h3 id="out">成员列表</h3>
+<p class="memberlist" style="line-height:2;"></p>
+
+<script src="https://apps.bdimg.com/libs/jquery/1.6.4/jquery.min.js" type="text/javascript"></script>
```
chrisguindon, line 19: This is not needed, we already load jquery with the site. When we need to use js on a page, we will define it in the {{ define "footer-suffix" }} region.

Example: https://github.com/EclipseFdn/iot.eclipse.org/blob/master/layouts/index.html

layouts/shortcodes/pages/jakartaeechina/index.zh.html
chrisguindon, line 25: Do we need to pull-in this data from an external domain?

laliwa: Since we can not reach the google form in China from time to time(just as now), we will need an online form registration.
Once someone has registered the form, we need to update his info on the members list on jakarta.ee/about/jakartachina page.
And because it is not so convenient to update the list timely, I think it is better to pull it from somewhere where things can be automately updated. For the time being , I deploy them on an external domain,  if these can be done on jakarta.ee, an external domain will not be needed.
Here I draw a picture( sorry for that I am not so good at it) to describe how it works. and due to my awful English, I'm not sure if I express everything clear. -_-
image

layouts/shortcodes/pages/jakartaeechina/index.zh.html
laliwa: Ok, got it.

layouts/shortcodes/pages/connect.zh.html
laliwa: I think I can have a try.

layouts/shortcodes/pages/jakartaeechina/index.zh.html
laliwa: Sure, I noticed that, and still need some time to work it out.

layouts/shortcodes/pages/connect.zh.html
autumnfound: I've started work on this, don't worry about it. I'll have a version up in the next hour or so. I had some meetings that delayed me having a version finished

laliwa: That will be fantastic, to be honest, I am not so familiar with the it. and I really afraid that I will make things mess.^-^

autumnfound:
@laliwa I've updated the shortcodes mentioned earlier to use i18n. We have only a couple of straggling values on the compatibility page, but it otherwise was straightforward!

layouts/shortcodes/pages/jakartaeechina/index.zh.html
laliwa: Hi Chris,I had tried this but it doesn't work.Pls do me a favor to help me to fix that. 
```
{{ define "footer-suffix" }}
<script>
(function($, document) {
	$.ajax({ 
			type:"get",
			dataType:"json",
			url: "https://jakarta.eaworld.io/getlisted/memberlist.php", 
			success: function(data){
				$.each(data,function(n,v){
					$.each(v,function(inn,ivv){
						$(".memberlist").append("*&nbsp;"+ivv+"<br>");
					});
				});
			}
	});
})(jQuery, document);
</script>
{{ end }}
```

laliwa:
Hi Chris，
there is something I'm a little confused with. about info from data\ dir e.g. data\membership\testimonials. i18n does not affect them, how sholud I translate this part?

autumnfound:
> Hi @chrisguindon ，
> there is something I'm a little confused with. about info from \data dir e.g. \data\membership\testimonials. i18n does not affect them, how sholud I translate this part?

We'd need to add multilingual support to that component, as right now it only supports a single file.

laliwa:
> > Hi @chrisguindon ，
> > there is something I'm a little confused with. about info from \data dir e.g. \data\membership\testimonials. i18n does not affect them, how sholud I translate this part?
> 
> We'd need to add multilingual support to that component, as right now it only supports a single file.

OK.

chrisguindon:
@autumnfound @laliwa I will create an issue under https://github.com/EclipseFdn/hugo-solstice-theme to request that we implement multilingual support for the testimonials component.

laliwa:
> @autumnfound @laliwa I will create an issue under https://github.com/EclipseFdn/hugo-solstice-theme to request that we implement multilingual support for the testimonials component.

@chrisguindon I'm afraid not only the testimonials but all that defined in /data dir which is refered by the pages, will need this work.

chrisguindon:
> > @autumnfound @laliwa I will create an issue under https://github.com/EclipseFdn/hugo-solstice-theme to request that we implement multilingual support for the testimonials component.
> 
> @chrisguindon I'm afraid not only the testimonials but all that defined in /data dir which is refered by the pages, will need this work.

Do we need to translate all of the data/content into Chinese? I was under the impressions that only key pages where going to be translated.

I think it's going to be hard to synchronize translations for pages like Compatible Products: https://jakarta.ee/compatibility/

laliwa:
> > > @autumnfound @laliwa I will create an issue under https://github.com/EclipseFdn/hugo-solstice-theme to request that we implement multilingual support for the testimonials component.
> > 
> > 
> > @chrisguindon I'm afraid not only the testimonials but all that defined in /data dir which is refered by the pages, will need this work.
> 
> Do we need to translate all of the data/content into Chinese? I was under the impression that only key pages where going to be translated.
> 
> I think it's going to be hard to synchronize translations for pages like Compatible Products: https://jakarta.ee/compatibility/

To be honest, for your first question, I'm not sure , I think that could be left for the committee to discuss.
As for Compatible pages and so on, I think only the title should be translated, but they are also defined in /data as yml file.

chrisguindon:
> To be honest, for your first question, I'm not sure , I think that could be left for the committee to discuss.
> As for Compatible pages and so on, I think only the title should be translated, but they are also defined in /data as yml file.

I agree. I think we are at a point where we need to define exactly which page should be translated since translating all our components is a significant effort for my team.

chrisguindon:
> To be honest, for your first question, I'm not sure , I think that could be left for the committee to discuss.
> As for Compatible pages and so on, I think only the title should be translated, but they are also defined in /data as yml file.

@shabnammayel @TanjaObradovic I think we need the MKT committee to discuss and define which section of the site must be translated.

For example, do we really want to translate testimonials and our list compatible products?

chrisguindon:
Closed in favor of https://github.com/jakartaee/jakarta.ee/pull/917

PR #916, opened 2020-10-15
Use summary60Char if defined, otherwise truncate summary to 57 characters

ivargrimstad:
Signed-off-by: Ivar Grimstad <ivar.grimstad@eclipse-foundation.org>

Applies to specification list

### DIFF
--- a/layouts/specifications/list.html
+++ b/layouts/specifications/list.html
@@ -31,10 +31,10 @@
       {{ range .Page.Sections.ByWeight}}
         <li>
           <h3><a href="{{- with .Params.redirect_url | default .RelPermalink }} {{.}} {{- end }}">{{ .Title }}</a></h3>
-          {{ if eq .Params.summary60Char nil  }}
+          {{ if eq .Params.summary_sixty_char nil  }}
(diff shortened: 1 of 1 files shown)


layouts/specifications/list.html
```diff
@@ -31,7 +31,11 @@ <h1>
       {{ range .Page.Sections.ByWeight}}
         <li>
           <h3><a href="{{- with .Params.redirect_url | default .RelPermalink }} {{.}} {{- end }}">{{ .Title }}</a></h3>
-          <p>{{ .Params.summary | truncate 57 }}</p>
+          {{ if eq .Params.summary60Char nil  }}
```
autumnfound, line 34: you can use `if isset .Params "summary60Char"` though I'm not sold on the name of this parameter, as it breaks the convention we currently follow for parameter names.

ivargrimstad: `if isset .Params "summary60Char"` doesn't work. Believe me, I tried... The problem is that the parameter does not need to be there, thus undefined (or nil).
Regarding the name, that's what was agreed upon when this was initially implemented (and somehow disappeared in a redesign of the page).

PR #922, opened 2020-10-22
Yiliu/src/only match height by row

flora8984461:
To fix https://github.com/EclipseFdn/hugo-solstice-theme/issues/175

and it's a combination with https://github.com/EclipseFdn/jquery-eclipsefdn-api/pull/30 and https://github.com/EclipseFdn/solstice-assets/pull/183

- [x] needs https://github.com/EclipseFdn/jquery-eclipsefdn-api/pull/30 merged before this
- [x] needs https://github.com/EclipseFdn/solstice-assets/pull/183 merged before this


flora8984461:
We can close this since the data-mh is added to default.

PR #931, opened 2020-10-29
update to Hugo 0.76.5

chrisguindon:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
 install:
 - npm install
 - npm run specifications
-- wget https://github.com/gohugoio/hugo/releases/download/v0.62.2/hugo_0.62.2_Linux-64bit.deb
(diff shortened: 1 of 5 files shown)


flora8984461:
Do we want to fix this warning too before merging or create a new issue?

image

chrisguindon:
> Do we want to fix this warning too before merging or create a new issue?
> 
> image

Where is the error coming from?

flora8984461:
> > Do we want to fix this warning too before merging or create a new issue?
> > image
> 
> Where is the warning coming from? It would be best to fix it now if we can

This is confusing, the warning does not tell me where the cause is.
I have checked all the `isset ` we are calling, while I don't think anything is wrong here. They all match the syntax:
isset COLLECTION INDEX / isset COLLECTION KEY  🤔

flora8984461 (approved):
lgtm

PR #950, opened 2020-11-14
Publishing mtg min for the Steering Cmte Oct 2020

TanjaObradovic:
Please merge.


data/meeting_minutes.yml
```diff
@@ -757,6 +757,30 @@ items:
           year: 2018
 
    steering_committee:
+          -
```
autumnfound, line 760: This line's spacing is not consistent with the other lines that denote a new meeting minute document (Line 764 and 768 are also examples of this). If you match the number of spaces on those lines on this one, it should pass. The storage format used for this is a bit dependent on spacing.

autumnfound (changes requested):

data/meeting_minutes.yml
TanjaObradovic: Should be all fixed now.

autumnfound (approved):
LGTM!

PR #946, opened 2020-11-12
SEO Optimization: Meta Refresh #941

flora8984461:
Part of #941 

------------- Updates  Nov 16-------------------
Specifications will not be in this PR
I'm not sure whether the preview can be searched by google, probably need to merge and test the FAQ schema? @chrisguindon 

-------------------------------------------------
Notes:
- Specifications not done since it's not built up on my localhost
- Compatible products: We need to add application server or platform to each product description- this part is on hold."
- About: Need another page/section to explain what Jakarta EE is
- specifications/platform/ not build on localhost
- /specifications/platform/8/platform-spec-8.html
- /specifications/platform/8/platform-spec-8.html   

My error while building specifications:

  - My static/specifications/ is empty
  - specifications/platform page is empty for me

```
> eclipsefdn-hugo-jakartaee@0.0.3 specifications C:\Users\yi\localhugo\jakarta.ee
> bash build.sh

build.sh: line 14: $'\r': command not found
Processing Jakarta EE Specifications...

build.sh: line 16: $'\r': command not found
Step 1: Clean up specifications content

build.sh: line 19: $'\r': command not found
Step 2: Clone specifications git repository
'...ing into 'content/specifications
remote: Enumerating objects: 70, done.
remote: Counting objects: 100% (70/70), done.
remote: Compressing objects: 100% (59/59), done.
remote: Total 74578 (delta 26), reused 38 (delta 11), pack-reused 74508
Receiving objects: 100% (74578/74578), 184.83 MiB | 9.20 MiB/s, done.
Resolving deltas: 100% (66594/66594), done.
Checking out files: 100% (32355/32355), done.
build.sh: line 22: $'\r': command not found

Step 3: Remove specifications/.git

build.sh: line 25: $'\r': command not found
Step 4: Create static/specifications folder
build.sh: line 28: cd: $'content\r': No such file or directory
Current working directory: /mnt/c/Users/yi/localhugo/jakarta.ee

build.sh: line 30: $'\r': command not found
Step 5: Preparing to move html files to static folder...
build.sh: line 32: syntax error near unexpected token `$'do\r''
'uild.sh: line 32: `for f in specifications/*/*/*.html; do
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! eclipsefdn-hugo-jakartaee@0.0.3 specifications: `bash build.sh`
npm ERR! Exit status 2
npm ERR!
[...]

Issue #941, opened 2020-11-11
SEO Optimization: Meta Refresh


excentrickristy:
Here are the new meta tags for the site  as a result of the SEO exercise: https://docs.google.com/spreadsheets/d/1ZQClxxCM1QWbjBKnqfrcrMIPUnYodhyEntDlZ0_vYW4/edit?usp=sharing

Any questions, please let me know! kristy@excentric.ca

aeiras:
Hola Kristy, 

Are you providing visibility on the SEO work on behalf of the Vendor hired to help with the activity?

excentrickristy:
Hi Amelia,
Yes, Excentric was the vendor and I was instructed to provide visibility here by Shabnam/Karen and @chrisguindon

aeiras:
Kristy, 

This issue and the form to use the Jakarta EE work tools are excellent and deeply appreciated. Thank you! 

On behalf of Tomitribe, I thank you for submitting your very first issue. So fantastic!  

Some of us care about stats & communicating effectively.  As such, we look forward to collaborating with you via open-channels!

PR #999, opened 2020-11-24
SEO Optimization: Meta Refresh #941

flora8984461:
Updates from https://github.com/jakartaee/jakarta.ee/pull/946 to remove messy commit history

**Note**: Specifications will not be in this PR


ericpoirier (approved):

data/faq.json
```diff
@@ -0,0 +1,110 @@
...
...E Trademark Guidelines.</p><p>At a high level, the use of the Compatibility logo is limited to only members of the Jakarta EE Working Group who are licensees under the Jakarta EE Trademark License Agreement or approved guest members of the Working Group (who are also licensees under the Eclipse Foundation Trademark License).</p><p>We are working with our friends at the Apache Software Foundation to allow their projects to use the Compatibility logo at no charge. We are open to working with other open source foundations that are interested in making the logo available to their eligible projects.</p><p>We require membership in the Working Group because the Jakarta EE Compatibility logo and brand have value to our community and the industry. We rely on membership funding to financially support the management of the Jakarta EE specification process, compatibility rules, and developer outreach activities to drive awareness and adoption. Also, Working Group membership fees fund the promotion of the Jakarta EE Compatible brand and its value in the marketplace.</p><p>Refer to <a href='https://jakarta.ee/legal/trademark_guidelines/'>Jakarta EE Trademark Guidelines</a> for more details.</p>"
```
chrisguindon, line 100: I think we should remove the A: from the text. I am concern that this will be included in the search results.

chrisguindon (changes requested):

data/faq.json
flora8984461: Emmm.. I'm actually hoping to keep the "A:" in https://jakarta.ee/about/faq/ , is it okay we remove them? If I add "A:" in the shortcode to all the answers, it will be weird to arrange the <p> and keep them in the same lines ... I'm still trying to find the best way to do so

chrisguindon: i was thinking that we should remove the A: completely...

flora8984461: Actually if the community agrees, we do not need "Q" and "A" in front of every question and answer 🤔

chrisguindon: +1 Let's make the change... I am pretty sure I was the one that added those A: in the first place..

chrisguindon: @TanjaObradovic Any objections with us removing the A: before each answer on the FAQ page? I think it would be best to remove it for the SEO work that we are doing!

TanjaObradovic:
No objections for removing both "Q:" and "A:" on the FAQ pages.

data/faq.json
flora8984461: removed.

PR #1000, opened 2020-11-24
try to fix npm error with windows

### LINKED ISSUE #620: npm install fails on windows

flora8984461:
Try to Fix #620 
--------------- Updates-----------------------
Removed bat since it's not needed.
Just add a new command to run for windows `"specifications_windows": "bash build.sh"`

There will be some errors when copying to static/, but most works fine. 
To fix the errors during copy to static/, you can use Notepad++ to open and set EOL Conversion to unix(LF)

--------------- Updates after a test to pull my branch -----------------------

Sadly I have to convert it again to unix(LF) by Notepad++ after pull a brand new directory😥
But this can be a quick workaround temporarily... 

---------------------------------------------------------------------------------

I need more time to investigate how to write a real script for windows batch,  but I found a workaround without doing so.

The new `build.bat` file I have uploaded is converted by my notepad++, here is what I did:
I copy the `build.sh` and rename to `build.bat`,
I use Notepad++ to open and set EOL Conversion to unix(LF),
Then I run the script `npm run specifications_windows` it works for me on Powershell

I'm not sure if it can still be successful after I upload to git and pull again as a new. I'm creating a draft PR to have a check.


Dexmaster:
You don't need all of this if you run bash first... (because your fix just runs bash) Also it can run original .sh file just fine...

autumnfound:
@flora8984461 does the fix mentioned above work? If it does, we could probably add bash inline to the commands in the package.json to make this platform agnostic

flora8984461:
> @flora8984461 does the fix mentioned above work? If it does, we could probably add bash inline to the commands in the package.json to make this platform agnostic

Yeah, just use bash can do the work. I'll remove the unnecessary file and update this PR

PR #1019, opened 2020-12-04
Update Jakarta EE Compatibility Page #1013

### LINKED ISSUE #1013: Update Jakarta EE Compatibility Page
On this page: https://jakarta.ee/compatibility/

We need to do the following:
- remove the orange Download button that is top right on the page
- add a new button the to left of "Compatibility Requests" that says "Download Compatible Products"
- when you click on that new button, it will take you to a new page
- the design file for this new page can be found here: https://drive.google.com/file/d/1IpRSyn_hOVcgxOk1AoQx8Ef7r7vVDOjT/view?ts=5fc8f5e1

flora8984461:
Fix #1013 

### DIFF
--- a/data/compatible_products.yml
+++ b/data/compatible_products.yml
@@ -12,8 +12,8 @@
 
 sets:
   - jakartaee_version: 9
-    items: 
-    - title: Jakarta EE 9 Platform Compatible Products
+    items:
+    - title: Jakarta EE 9 Platform
       items:
       - name: Eclipse GlassFish
         vendor: Eclipse Foundation
@@ -22,7 +22,7 @@ sets:
         versions:
         - version: 6.0
           compatibility: 'https://glassfish.org/certifications/jakarta-full-profile/9.0/TCK-Results'
-    - title: Jakarta EE 9 Web Profile Compatible Products
+    - title: Jakarta EE 9 Web Profile
       items:
       - name: Eclipse GlassFish
         vendor: Eclipse Foundation
@@ -34,7 +34,7 @@ sets:
 
   - jakartaee_version: 8
     items:
-    - title: Jakarta EE 8 Platform Compatible Products
+    - title: Jakarta EE 8 Platform
       items:
       - name: Payara Server 
         vendor: Payara Services Limited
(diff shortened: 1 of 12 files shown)


data/compatible_products.yml
```diff
@@ -121,6 +142,7 @@ sets:
         versions:
         - version: 5.1.0
           compatibility: 'https://eclipse-ee4j.github.io/glassfish/certifications/jakarta-web-profile/8.0/TCK-Results'
+          version_download: 'https://eclipse-ee4j.github.io/glassfish/certifications/jakarta-web-profile/8.0/TCK-Results'
```
chrisguindon, line 145: "version_download" seems redundant here. Can we rename this to "download_url" instead of "version_download"?

less/styles.less
```diff
@@ -50,6 +50,10 @@
   font-size: 100px;
 }
 
+.font-roboto {
```
chrisguindon, line 53: I don't think we should be introducing a new font here. Why are we not using the website font default?

chrisguindon (changes requested):

less/styles.less
flora8984461: I see the design mockup is using roboto, I thought they want to use it? 🤔

If not, I'm gonna remove

flora8984461: I find it probably due to the design tool, it seems automatically use this font... 🤔

flora8984461: removed

data/compatible_products.yml
flora8984461: renamed

layouts/shortcodes/pages/compatibility-products.html
```diff
@@ -1,18 +1,9 @@
 <div class="row">
-  <div class="col-md-16">{{ i18n "compatibility-intro-text" | safeHTML }}</div>
-  <div class="col-md-8"><img class="compatible-logo img img-responsive" src="/images/jakarta/jakarta-ee-compatible-logo-color.svg" title="{{ i18n "jakarta-ee-logo-alt" }}"></div>
+  <div class="{{ with .Page.Params.compatibility_download_wrapper }}{{ . }}{{ end }}" style="text-align: center;">
+    {{ i18n "compatibility-download-button" | safeHTML }}
+    <a class="btn btn-lg btn-secondary margin-bottom-5" href="https://githubcom/eclipse-ee4j/jakartaee-platform/issues?q=is%3Aissue+is%3Aclosed+label%3Acertification">Compatibility Requests</a></div>
```
flora8984461, line 4: I'm doing this `compatibility_download_wrapper` because I added Chinese "compatibility-download-button", and in Chinese, the button is shorter, I need to keep the button align right for both Zh and En pages.

Issue #1052, opened 2021-01-18
[IE Compatibility] Only IE: Contributor card not showing


flora8984461:
Updates:

Compatible Products display is fixed on IE.

Only Contributor card not showing for IE remains. (Need to modify the syntax to drop syntax that not supported by IE.)

--------------------

For release page:

with err: `SCRIPT438: Object doesn't support property or method 'forEach'`
![image](https://user-images.githubusercontent.com/39588094/105895953-3289b580-5fe4-11eb-8987-c8090178eaf4.png)

chrisguindon:
This is bad... let's plan at looking at this on Monday!

Great find @flora8984461 !

oliviergoulet5:
I can't reproduce this on IE11 mode for Edge. It looks like it has already been fixed.

![image](https://user-images.githubusercontent.com/17685861/212919412-720a5ace-6ebf-484c-a9dc-b34f95bc1e62.png)

PR #1060, opened 2021-01-26
Added width to fix IE image distorted issue #1052

flora8984461:
Part of #1052 

For Compatibility Page, after the fix:

image

image

Might still be a little bit not perfect on IE, but at least not distorted UI. 🤔 (I think we don't want to spend a lot on IE)

### DIFF
--- a/js/contributor-cards/index.js
+++ b/js/contributor-cards/index.js
@@ -5,6 +5,10 @@ import contributorList from '../../data/contributors/jakarta-ee-9.json';
 
 const cardList = document.querySelector('.card-list.contributor-list');
 
+if (window.NodeList && !NodeList.prototype.forEach) {
+  NodeList.prototype.forEach = Array.prototype.forEach;
+}
+
 if (!!cardList) {
   // at least login should exist
   const filteredList = contributorList.filter(({ login }) => !!login);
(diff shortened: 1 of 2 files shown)


autumnfound (approved):

flora8984461:
@autumnfound Added one more fix, re-request you > <

js/contributor-cards/index.js
```diff
@@ -5,6 +5,10 @@ import contributorList from '../../data/contributors/jakarta-ee-9.json';
 
 const cardList = document.querySelector('.card-list.contributor-list');
 
+if (window.NodeList && !NodeList.prototype.forEach) {
+  NodeList.prototype.forEach = Array.prototype.forEach;
+}
+
```
flora8984461: @autumnfound Is this snippet also not allowed? this is from https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach 
If so, I will remove it and make this PR fix IE images first. And find a way to add polyfill later.

autumnfound: Yeah we should remove that snippet. Once that is done we can look at merging this

flora8984461: removed

chrisguindon:
@flora8984461 Can you re-roll this PR. There is now a conflict now that I merged #1061

chrisguindon (approved):

flora8984461:
conflict fixed @chrisguindon

PR #1072, opened 2021-02-19
meeting minutes Dec10, Jan07, Jan14, and Jan28

kazumura:
Signed-off-by: Kenji Kazumura <kzr@fujitsu.com>


autumnfound:
To make the new meeting minutes, we need to add them as entries in the data file for meeting minutes in the right section (order is important, as it doesn't reorder the file). The file is at https://github.com/kazumura/jakarta.ee/blob/min-dec-jan/data/meeting_minutes.yml. Let me know if you need some help with this!

autumnfound (approved):

PR #1083, opened 2021-03-04
update internal doc refs to get to the proper section at least

kwsutter:
Signed-off-by: Kevin Sutter <kwsutter@gmail.com>

The markdown rendering in Hugo didn't accomplish what I was looking for.  I was trying to create a reference for a specific step in the process, but I've tried several different variations and can't get it to work properly.  So, I'm going to revert to just mentioning the desired steps and reference the overall section on creating a final specification.  Good enough...

### DIFF
--- a/content/committees/specification/operations/index.md
+++ b/content/committees/specification/operations/index.md
@@ -248,7 +248,7 @@ Although a release review is not required, several of the following requirements
 A release review will have validated that the specification
 project has:
 
-1.  <a name="staging_artifacts">Produced</a> a final staging release via OSSRH staging repository for the api,
+1.  Produced a final staging release via OSSRH staging repository for the api,
     javadoc. 
 
     -   Javadocs should include the [Eclipse Foundation Specification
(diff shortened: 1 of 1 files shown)


kwsutter:
@ivargrimstad or @chrisguindon, if you could take a quick review and merge of this PR, I would appreciate it.  I'm open to other alternatives, but I haven't been able to get this to work as I had hoped.  At a minimum, this PR will at least get the reader back to the "create final spec" section.  Thanks!

ivargrimstad (approved):

autumnfound:
I'm in the process of making a PR that addresses this issue in our base theme. Adding an ID to a document should be trivial but it really isn't in most cases, which is one of the gaps in Hugo for sure! It'll be as easy as this in the future:

`{{<text_target id="sample_heading">}}Some text here to add an ID to{{</ text_target >}}`

kwsutter:
> I'm in the process of making a PR that addresses this issue in our base theme. Adding an ID to a document should be trivial but it really isn't in most cases, which is one of the gaps in Hugo for sure! It'll be as easy as this in the future:
> 
> `{{<text_target id="sample_heading">}}Some text here to add an ID to{{</ text_target >}}`

That will be excellent!  Thanks!  Can you link the PR here or send it to my via Slack?  I'd like to know when it becomes real.

autumnfound:
@kwsutter https://github.com/EclipseFdn/hugo-solstice-theme/pull/205 is the PR! :)

PR #1084, opened 2021-03-04
Updated Hugo theme version to latest for text target component access

autumnfound:


autumnfound:
@kwsutter Once this is merged in, you should have access to the new text target component I mentioned earlier today! It can go right in markdown, so it should be easy to drop in. There's a className attribute in case you want to add a class to the span to bold it for example (not required, but just letting you know).

PR #1127, opened 2021-04-29
Added Istanbul JUG into guest members list

linkfang:


TanjaObradovic:
Looks great! Thanks!
Can we also have Participant Members and Guest Members section with the same font size as Strategic and Enterprise Members section please?

ivargrimstad:
> Looks great! Thanks!
> Can we also have Participant Members and Guest Members section with the same font size as Strategic and Enterprise Members section please?

I think the font sizes were intentional to amplify the membership levels. I'll merge this one, we can take the font adjustments in separate PR if we go there...

PR #1128, opened 2021-05-03
Added wildfly preview into compatible list-for issue1126

linkfang:


ericpoirier:
Not sure if that's intended or not but WildFly is listed twice.

Screen Shot 2021-05-03 at 11 25 50 AM

I know that one is a preview and the other one is not but just want to make sure it is intended.

ericpoirier (approved):

linkfang:
As requested, changed the WildFly Preview to WildFly, and moved the preview to the version number "Preview 23.0.1.Final".
So now we only have 1 WildFly in the list instead of 2.

autumnfound (changes requested):
> As long as we have **both** versions listed for WildFly, 23.0.0 and Preview 23.0.1.Final that would be fine. If I am missing something please let me know. @bstansberry @scottmarlow can you please confirm?

We shouldn't be removing the already existing v23 for the Web compat section. We should have both (as both are considered current versions for the web profile)

linkfang:
Hi @autumnfound , 

The message you quoted was Tanja trying to confirm with bstansberry and scottmarlow, and they both gave a +1 based on current change. (Only has 1 preview version instead of 2)

So, I am not sure, should we confirm with bstansberry and scottmarlow again?

bstansberry:
@linkfang @autumnfound AIUI, https://deploy-preview-1128--jakartaee.netlify.app/compatibility/ is the result of the current state of this PR. In that case +1 to this change.

autumnfound:
> @linkfang @autumnfound AIUI, https://deploy-preview-1128--jakartaee.netlify.app/compatibility/ is the result of the current state of this PR. In that case +1 to this change.

If this is as expected for you guys, then we can merge it! I just wanted to try and make sure both the stable and preview were available as versions (as v 23 would be removed by this patch)

bstansberry:
@linkfang Yes, both the standard WildFly and the WildFly Preview downloads are available from https://www.wildfly.org/downloads/. So this is good to merge.

autumnfound (approved):

PR #1130, opened 2021-05-04
Clean news section

### LINKED ISSUE #1129: Remove test news articles
A lot of news items contains "lorem ipsum..." text

ivargrimstad:
Will close #1129


autumnfound:
-1 on this change set. While I agree we should remove text that is placeholder, we should check if these articles are fake or real content, and whether we have real content copies for them if they are real. These files feed the announcements section on the home page, and removing the entire files would leave that section empty.

autumnfound (changes requested):
see comment above

ivargrimstad:
> -1 on this change set. While I agree we should remove text that is placeholder, we should check if these articles are fake or real content, and whether we have real content copies for them if they are real. These files feed the announcements section on the home page, and removing the entire files would leave that section empty.

I only removed the ones that were just created for test purposes. Having a bunch of news items with wrong banners and "lorem ipsum..." does not look good at all.
The rest of the items that make sense are still there, so the section will not be empty.

TanjaObradovic (approved):
@autumnfound I agree with @ivargrimstad please delete all articles with no text / content from the list.

autumnfound:
> @autumnfound I agree with @ivargrimstad please delete all articles with no text / content from the list.

The big reason for the minus 1 was to create conversation before it merged to make sure it was all test content and none of it was real content with fluff added in. If that's the case I can merge it today!

Issue #1164, opened 2021-05-11
Add The Apache Software Foundation as Guest Member of the WG


ZahraFazli:
Hello 
Please add The Apache Software Foundation as Guest Member of the working group.

[asf_logo_url.pdf](https://github.com/jakartaee/jakarta.ee/files/6461649/asf_logo_url.pdf)

linkfang:
Hi @ZahraFazli , The Apache Software Foundation is added, please review here:
https://deploy-preview-1171--jakartaee.netlify.app/membership/members/

ZahraFazli:
Looks good, thanks

PR #1171, opened 2021-05-13
Added Apache as a guest member

linkfang:
[Issue1164](https://github.com/jakartaee/jakarta.ee/issues/1164)


layouts/shortcodes/pages/membership/explore-members.html
```diff
@@ -41,6 +41,7 @@ <h3 class="heading-line text-center">
           <span>{{ i18n "members-guest-members-title" }}</span>
         </h3>
         <ul class="list-inline text-center">
+          <li><a href="https://www.apache.org/" target="_blank"><img src="/images/members/jakarta_member_apache.png" width="115" class="img-responsive"></a></li>
```
autumnfound: I'd recommend using width 145 here. While we want to keep them small, we want them to be comparable in size and legible

linkfang: OK, cool.
I tried 145 and 130, I think 130 might be better. 145 looks a little big compared to others.
What do you think?

130:
image

145:
image

autumnfound (approved):

linkfang:
Got a +1 from Zahra, good to merge after code review

Issue #1026, opened 2020-12-08
Improve Netlify Build Time


chrisguindon:
The average build time for jakarta.ee with Netlify is about 6 minutes. I think we can do better here.

Some ideas on how we can solve this problem:

- We are cloning the specification repository for each build. Can we improve the build time by including the specification repo as a sub-module?

- Do we need to run npm install at the start of each build?

PR #1182, opened 2021-05-18
Improve Netlify Build Time #1026

chrisguindon:

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "watch": "NODE_ENV=development webpack --watch --progress --config=node_modules/laravel-mix/setup/webpack.config.js",
     "hot": "NODE_ENV=development webpack-dev-server --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
     "production": "NODE_ENV=production webpack --progress --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "install": "NODE_ENV=production npm run production",
     "postinstall": "NODE_ENV=production npm run fetch_project",
     "specifications": "bash build.sh",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "install": "NODE_ENV=production npm run production",
     "postinstall": "NODE_ENV=production npm run fetch_project",
     "specifications": "bash build.sh",
-    "build_netlify": "NODE_ENV=production npm run production && npm run specifications && npm run fetch_project",
+    "build_netlify": "NODE_ENV=production npm run production && npm run fetch_project",
```
autumnfound, line 22: If we update this in conjunction with the redirect you have, then once this hits production level, this will no longer work as all of the netlify builds use this command. Maybe we should split it into `build_netlify_light` and `build_netlify`?

chrisguindon: Why would this no longer work?

chrisguindon: I think you wrote your comment while I was still working on this.

I since changed the URL to be a relative path. The proxy works on previews and production now.

The build time went from ~5-6 minutes to 1 minute!

autumnfound: Once this hits production, nothing would ever build the specs, which is what the redirects rely on

chrisguindon: Yeah, that is the goal! We no longer need to build the specification section for this website. It will use the content from https://github.com/jakartaee/specifications directly!

autumnfound: Ahh nvm, Christopher reached out and pointed out these are different sites. D'oh!

autumnfound:
Did we want to merge this into src or redesign?

chrisguindon:
@autumnfound I will merge this to src. Feel free to merge it in the redesign branch after!

PR #1183, opened 2021-05-18
Restructure LESS to split out special page Less

autumnfound:
Moving CSS around to simplify the process of the redesign. Added variables file to override current base look and feel. This will need to stay separate from current base look and feel in solstice assets as multiple sites rely on the current look and feel.


less/compatibility.less
```diff
@@ -0,0 +1,202 @@
...
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * Contributors:
+ *   Andrii Malytskyi <amalytskyi@tomitribe.com>
+ *   Jonathan Gallimore <jgallimore@tomitribe.com>
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+*/
+
+.d-flex {
```
ericpoirier, line 15: Would it make more sense to create a class called .display-flex instead of d-flex?

Not sure its super obvious "d" means display here.

ericpoirier: Also would it be worth it to export that to solstice-assets? There's a few classes here that I think we could add to solstice-assets

autumnfound: d is for display, and its a bootstrap 4 > thing. https://getbootstrap.com/docs/5.0/utilities/display/ We're going to be eventually re-writing this entire file, so there isn't really a point for the compatibility file

PR #1192, opened 2021-05-22
Update Page with Soujava as adopters

otaviojava:


ivargrimstad (approved):
For some reason, the preview doesn't work...
But the changes look ok to me.

PR #1194, opened 2021-05-24
Update 9.1 committers list

Dexmaster:
Previous version was before end (18 May) this is today generated version.

Plus fixed issue with multi-author commits not contributing both authors. (now up to 10 authors per commit are included)
Also fix for pageTraversing was not working for commits.

P.S. As a result 2 new committers in a list.


aeiras:
Thank you, Andrii for looking at the code & helping freeze the version to match the 5/18 date + more details 💙

a total count of 25 contributors not including those behind the scenes. 🥋

PR #1207, opened 2021-05-28
"Adopt a Spec" Program for JUGs

gencdevops:


gencdevops:
eclipsefdn / eca  I accepted the contract. But I'm still getting error

autumnfound:
> eclipsefdn / eca I accepted the contract. But I'm still getting error

The ECA also requires you to set your Github handle into your Eclipse account so we can match to an account properly. Once that's done as well, you should be able to "revalidate" after clicking the details link next to the ECA check in the checks screen

gencdevops:
when i click the details I get the following error. 
Taner Genç (67705077+gencdevops@****.noreply) is not covered by the necessary legal agreements in order to proceed.

PR #1256, opened 2021-09-08
Small updates

ericpoirier:


less/redesign.less
```diff
@@ -40,6 +40,11 @@ footer#solstice-footer {
     background-repeat: no-repeat;
     padding-bottom: 0px;
   }
+  h2 {
+    img {
```
autumnfound: we should combine these

less/components/_commons.less
```diff
@@ -188,4 +191,12 @@ footer#solstice-footer {
...
     display: flex; align-items: center;
-}
\ No newline at end of file
+}
+.featured-section-members {
+    .owl-item {
+        display: flex;
+        justify-content: flex-end;
+        flex-direction: column;
+        height:115px;
```
autumnfound: our style guide suggests spaces between the colon and the value. There are a few other instances of this, but we can put this off until your final cleanup of the CSS if you like

less/components/_commons.less
```diff
@@ -188,4 +191,12 @@ footer#solstice-footer {
 
 .product-list .grid-center { 
     display: flex; align-items: center;
-}
\ No newline at end of file
+}
+.featured-section-members {
+    .owl-item {
```
autumnfound: we can combine these 2 selectors

autumnfound (commented):
we should remove the package-lock in favor of the yarn lock file

PR #1274, opened 2021-10-04
Fix mobile menu

ericpoirier:


less/components/menu.less
```diff
@@ -60,6 +60,55 @@
         padding-bottom:0;
     }
 }
+
+@media (max-width: @screen-xs-max) {
+    #main-menu {
+        .dropdown-menu {
```
autumnfound: We can probably combine this line with the previous as there is no classes that target the main-menu other than the ones under dropdown-menu.

autumnfound (changes requested):
The menu breaks when the page is resized after load.

Additionally, I'm not a fan of breaking the mobile first design, but I understand the reasoning here. It would have been an absolute bear to update all of the rules within the current structure, even if it would have been cleaner.

ericpoirier:
> The menu breaks when the page is resized after load.

It was breaking because of the match height that was implemented for desktop view.

I'm removing the match height now if the screen is small then 768px which resolves the issue.

> 
> Additionally, I'm not a fan of breaking the mobile first design, but I understand the reasoning here. It would have been an absolute bear to update all of the rules within the current structure, even if it would have been cleaner.

autumnfound (approved):

linkfang (approved):

PR #1282, opened 2021-10-08
Update the layout of the FAQ page

### LINKED ISSUE #1141: FAQ page refresh
The FAQ section should be updated to use a [Bootstrap accordion collapse](https://getbootstrap.com/docs/3.3/javascript/#collapse) section. Some minor JS may be needed to transition between the open and close arrows on the right side of the collapse block. This would bind to the `show.bs.collapse` and `hide.bs.collapse` events on the page, and would use some animation to transition between the 2 glyphs.

The simpler alternative would be to use 2 glyphs in the section that are hidden by CSS depending on the collapse state of the container. This would be simpler and more robust than the JS solution. In regards to choosing an implemention, let's use some timeboxed dev time (say 2-3 hours) to see if you can get the JS solution working with some degree of  stability, otherwise lets add the simpler solution. We can always upgrade this in the future!

Design:
https://projects.invisionapp.com/share/AX10M9QL74RN#/screens/448585117
image

ericpoirier:
#1141 


layouts/shortcodes/faq.html
```diff
@@ -0,0 +1,24 @@
+{{ $data := index .Site.Data .Site.Language.Lang }}
+
+<div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
+  
+  {{ range $data.faq.mainEntity }}
+    <div class="panel panel-default">
+      <div class="panel-heading" role="tab" id="heading-{{ .name | replaceRE "[^a-zA-Z]" "" | safeHTML }}">
```
autumnfound: why are we using safe HTML here? That seems like it introduces holes to do some really strange stuff here. Also, does urlize not remove the needed characters?

less/components/panels.less
```diff
@@ -0,0 +1,55 @@
+.panel-default {
```
autumnfound: is panel-default a custom class, or is it one defined by bootstrap for panel groups? If its not a custom class, we should consider namespacing these changes to not have styles leak into other components

less/components/panels.less
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+        a[aria-expanded="false"] {
+            .fa-chevron-right {
+                display: block;
+            }
+        }
+    }
+}
+
```
autumnfound, line 45: Is it intended to apply these styles to all panels? We use panels in a few ways, so we should be careful with a global override like this

layouts/shortcodes/faq.html
ericpoirier: Ok i'll use urlize then, much simpler solution. I just tested and it works.

less/components/panels.less
ericpoirier: I created the class .panel-bordered so that it doesn't take over the default styles.

less/components/panels.less
ericpoirier: By creating a custom class, these styles will only apply to elements that uses that class.

ericpoirier:
Got +1 from @shabnammayel

less/components/panels.less
autumnfound: These panels are outside the panel-bordered namespace atm, which is why I mention this

autumnfound (approved):

Issue #1146, opened 2021-05-06
Meeting minutes refresh


autumnfound:
The meeting minutes page is very similar to what is currently up on production, with a few small changes in looks but not functionality. The tab headings used on the homepage for the news and events sections will be used here as well, replacing the standard simple boxed design. Instead of surrounding the section in a border, each year will be surrounded with a box with a left-side highlight mark with the year contained.

Design:
https://projects.invisionapp.com/share/AX10M9QL74RN#/screens/448585127
![image](https://user-images.githubusercontent.com/23018241/117348491-3d23d880-ae78-11eb-8707-253d408c864d.png)

ericpoirier:
@shabnammayel I created a PR that updates the meeting minutes page.

You can see the preview here:
https://deploy-preview-1300--jakartaee.netlify.app/about/meeting_minutes/

Let me know if you'd like to change anything.

Otherwise, i'll wait for your +1 to merge to production.

Thanks.

shabnammayel:
+1 LGTM! Thanks

TanjaObradovic:
Looks great!

ericpoirier:
My changes have been merged to production.

PR #1300, opened 2021-10-28
issue-1146 - Update layout of Meeting Minutes page

ericpoirier:
#1146 


ericpoirier:
We have +1 from Shabnam.

less/components/blocks.less
```diff
@@ -25,4 +50,30 @@
     &:hover {
         color: darken(@bordered-box-link-color, 15%);
     }
+}
+
+.flag-box {
```
autumnfound: Couldn't we have used what we have for cards here? https://eclipsefdn-hugo-solstice-theme.netlify.app/components/cards/ all we need to do to align it properly is add a class that unsets left position, and sets right using whats in the left position

ericpoirier: Didn't know that existed, I'll using that instead. Thx.

chrisguindon (changes requested):
We need to add a redirect rule to avoid breaking any existing link: https://github.com/jakartaee/jakarta.ee/blob/src/static/_redirects

ericpoirier:
> We need to add a redirect rule to avoid breaking any existing link: https://github.com/jakartaee/jakarta.ee/blob/src/static/_redirects

I added the redirect to my PR.

chrisguindon (approved):

PR #1304, opened 2021-11-05
Issue 1151 - Update Newsletter page

### LINKED ISSUE #1151: Newsletter archive refresh
Newsletter archives is a simple facelift and rearranging of content to make use of the same look and feel that exists across the site refresh. The year header has a glyph to the left of it, and underneath the header is a simple list of links. Instead of being a full width section, these lists will be in smaller bordered boxes
 
@karenmcnaughton the presence of links is only assumed here as there is no indication that the text in these sections is linked. The best practice is to somehow decorate the text to indicate it isn't standard text to indicate the presence of a link. Either with different color text, an underline, or a glyph. 

The default would be to allow the standard treatment of links to apply, which would make the text a different color. Do you have a preference for some other look, or does this work for you?

Design:
https://projects.invisionapp.com/share/AX10M9QL74RN#/screens/448585128
image

ericpoirier:
#1151 


ericpoirier:
This is now ready to be reviewed and merged. Thanks.

layouts/shortcodes/pages/newsletter.html
```diff
@@ -1,189 +1,42 @@
...
-        </div>
-      </div>
+<h2 class="big-text-secondary margin-bottom-20">Jakarta EE Newsletters Archives</h2>
+
+<div class="row margin-bottom-40">
+  <div class="col-sm-8 card-container">
+    <div class="card-panel bordered panel panel-default match-height-item-by-row">
+      <h3 class="big-text text-secondary"><i class="fa fa-calendar-o" aria-hidden="true"></i> 2020</h3>
+      <ul>
+        <li>May: <a href="https://www.eclipse.org/community/eclipse_newsletter/2020/may/" target="_blank">Jakarta EE 9 and Beyond</a></li>
```
autumnfound: We should consider making the month clickable too just for convenience sake. WDYT?

ericpoirier: Sure, probably better to make the whole sentence clickable.

autumnfound (approved):

PR #1309, opened 2021-11-08
Redesign Working Group page

### LINKED ISSUE #1142: About Working Groups page refresh
The creation of this section should be a pretty straightforward process. The first 2 sections are segmented text sections, with the third being a more complex use case of the new cards component we are in the process of creating. This means that we will need to be able to configure additional classes per card in the data file so that we can mix colours, and also define type by entry rather than for all cards of a section (which is the more straightforward way).

Design:
https://projects.invisionapp.com/share/AX10M9QL74RN#/screens/448585113
image

ericpoirier:
#1142 #1324


ericpoirier:
We got the +1.

chrisguindon (approved):

autumnfound (approved):

ericpoirier:
We have the +1 from Shabnam. Merging!

PR #1318, opened 2021-11-11
Issue 1154 - Create Videos section

### LINKED ISSUE #1154: Resource videos refresh
The videos section refresh has 3 pieces of work to consider, 2 of which are on the critical path for the refresh, and the third being a new feature that allows for live filtering of the videos by video category. The top banner is a custom section using an image background and 3 visible sections of content in a 15-65-20 split. The first section containing a PNG logo in a container with a white bg and border radius. The second section is text with a couple of glyphs. The third section is a simple button.

The second component is more image + text sections displaying the list of videos defined in the site. A similar concept is done on the IoT site under the [Video Resources section](https://iot.eclipse.org/community/resources/videos/). We could use that implementation as a base with an updated look and feel. Currently, there are no images associated with videos, and it would be a task to have images be given on the addition of new videos. 

The new feature, which will be defined in a further issue scoped to post-launch changes would allow for the videos to be filtered by updating a dropdown that contains all of the different categories of videos present on the page.

Design:
https://projects.invisionapp.com/share/AX10M9QL74RN#/screens/448585122
image

ericpoirier:


layouts/shortcodes/pages/community/videos.html
```diff
@@ -0,0 +1,50 @@
+<div class="card-banner card-banner-dark padding-15 margin-bottom-60">
+    <div class="row">
+        <div class="col-sm-4 match-height-item-by-row vertical-align-container">
+            <span class="featured-circle-white">
+                <img src="/images/jakarta_ee_logo_stacked.svg" alt="Jakarta EE">
+            </span>
+        </div>
+        <div class="col-sm-12 match-height-item-by-row">
+            <h2 class="big-text">Subscribe to our Youtube Channel</h2>
```
autumnfound, line 9: We should probably set up these values to be i18n. I know I'm late to the party to start suggesting this, but better late then never. There was significant work to set these pages up to be compatible with language localization, and we should do our best to not take a step backwards

ericpoirier: I agree but english texts are changing. I was thinking that when we get to translate the content, we can then create the i18n variables.

ericpoirier:
We have the +1 from Shabnam.

autumnfound (approved):

autumnfound:
I don't necessarily agree about the i18n labels, but its better to get this stuff in than sweat the small stuff

Issue #1343, opened 2021-12-03
Root URL of the Specifciations List


ivargrimstad:
**Describe the bug**
The root URL of the Specification List should be https://jakarta.ee/specifications, not https://jakarta.ee/specifications/browse

- Since the beginning,  /specifications/ has been the location we communicate as the location of the specifications. 
- An extra click to get to the list is unnecessary
- It complicates the navigation back and forth between the specifications (e.g. clicking on the "Specifications" breadcrumb brings you to the root and you have to click browse again to get to the list.
- Browse is treated as a spec (#1342) in the list

ericpoirier:
@ivargrimstad we decided to put the "Overview" page on /specifications top keep things consistent with the other sections of the site.

If we were to put the Specifications List page back to /specifications it would break consitency with other sections.

I don't mind doing the change but I'll leave the decision between you and @shabnammayel.

Also if we are to make the change, what would be the url of the "Overview" page?

Thanks.

ivargrimstad:
> Also if we are to make the change, what would be the url of the "Overview" page?

Anywhere, something else than /specifications/
Then you won't need the special treatment of the browse page to get it out of the list (#1342)

shabnammayel:
@ericpoirier the reason we are creating an overview page is to anchor all external links in Jakarta EE's domain first and drive the traffic to our own page before going to external pages. So to solve this issue, we should move the Specification Projects to the "Resource dropdown" (we can use the same box that we created on the Specification Overview" page currently for the box but in the Resource page instead). Once we do that, all the other links on the Specification dropdown are internal and we won't need an overview page anymore. The Browse Specification URL can be https://jakarta.ee/specifications then. thanks

shabnammayel:
@excentrickristy looks like we don't have an icon for Specification Projects to move it to the "resource page". I was looking at the icons we have to see if we can reuse one of them but it looks like we have used them all in the resource page already! :) Can you please help with that?

ericpoirier:
@shabnammayel @ivargrimstad please review the following staging page and let me know if you'd like to make more changes.

https://deploy-preview-1345--jakartaee.netlify.app/specifications/

Thanks.

shabnammayel:
@ivargrimstad We will keep the name of the button "Browse Specification" instead of "Specification List" in the dropdown. But the URL will be the same as Eric has shared. WDYT?

ivargrimstad:
Yes, great! Thanks!

[...]

PR #1344, opened 2021-12-03
Remove Browse Specifications from list

ericpoirier:
#1343 


ivargrimstad:
This special treatment of the brows page won't be needed when #1343 is fixed, so you can hold off this a little...

PR #1350, opened 2021-12-06
Update layout of Specifications section

### LINKED ISSUE #1153: Specifications list page refresh
The list page has 3 new aspects to it compared to the current specification list page. The first is that the platform spec projects have been separated from the rest of the specifications. We can likely accomplish this by creating a string list of projects we want to be in the top section and just loop through the list of specs twice. The first time we will only print the projects that match the allowlist, and the second loop through we will print the rest of the specifications, using the before mentioned list as an exclude list.

The specification list page makes use of a lot of content that is already available on the page that is current, just not used, so a lot of the current logic can be preserved in terms of looping through specs which should save some time. Additionally, the buttons used to link to the project page and website exist on the specification summary page, as well as logic to fetch the implementations of the specification. Example spec page with buttons and implementation links: https://jakarta.ee/specifications/platform/

The second and more complex part of this page is a new filtration system. As this is the second time we've had to make use of filtration and flags, we should consider looking into a JS library to help us accomplish this such as [Fuze.js](https://fusejs.io/demo.html) (which falls under the APL 2 license). Fuse seems to support adding additional flags by value match, so we should be able to easily add search filters as well as text search with this product.

@chrisguindon WDYT of adding Fuse? We could add it just on pages where we need special text search, and it would greatly simplify our life now and in the future when we need to do stuff like this.

Design: 
https://projects.invisionapp.com/share/AX10M9QL74RN#/screens/448585120
image

ericpoirier:


chrisguindon:
@ericpoirier I see that you are modifying the build script. You will need to make sure that your changes doesn't break previews under the specifications repo: https://github.com/jakartaee/specifications

Whenever this is merged, you will want to submit a PR to the spec repo to make sure all is well.

ericpoirier:
We have the +1 from Shabnam

linkfang (approved):

chrisguindon (approved):

Issue #1391, opened 2022-02-01
Please update Payara alternate for Jakarta EE Steering Committee 


TanjaObradovic:
On the page https://jakarta.ee/committees/steering/
under 
_Committee Members_

Please replace Eliot Martin with  Hadar Vorenshtein 

It should look like this
Steve Millidge - Payara, Hadar Vorenshtein - alternate

ericpoirier:
Hi @TanjaObradovic, please review the following changes:

https://deploy-preview-1394--jakartaee.netlify.app/committees/steering/

Thanks.

TanjaObradovic:
Looks great!

ericpoirier:
This was merged.

Thanks.

PR #1394, opened 2022-02-02
Update Committee Member name

ericpoirier:
#1391 


ericpoirier:
We have the +1 from Tanja.

Issue #1401, opened 2022-02-08
Page "https://jugpe.com.br/" doesn't exist: Request to send us the right URL of "JUG PE".


DmitriGit:
Hello,

1) on the page "[https://jakarta.ee/specifications/jsonp/](https://urldefense.com/v3/__https:/jakarta.ee/specifications/jsonp/__;!!ACWV5N9M2RV99hQ!bx4hE0SK8SQPNEwhO4GxkGNXJul6mXaq8R2Vok-vgsN5sxFN52a6XcPc75glYi-UT7SX$)" ("Jakarta EE JSON Processing" project's page) a link to "JUG PE" (see the image "JUG PE") point to a site that doesn't exists - I have "Page doesn't exists" when click on it,

2) the same problem is also on the "https://jakarta.ee/community/adopt-a-spec/" page: if you click on "JUG PE" image, then will be displayed "Page doesn't exists".

Can "JUG PE" site send us the right URL ?

Thank you,
Dmitri.

ericpoirier:
@shabnammayel would you know how we can get the proper url for JUG PE?

Thanks.

ivargrimstad:
Maybe @vinyfalcao can help?

TanjaObradovic:
Until we hear from  @vinyfalcao  or maybe @otaviojava, we can use this link https://www.linkedin.com/in/jug-pe-713ab1191/?originalSubdomain=br

linkfang:
The link has been set to the LinkedIn page for now, please preview the change here:

https://deploy-preview-1405--jakartaee.netlify.app/specifications/jsonp/
https://deploy-preview-1405--jakartaee.netlify.app/community/adopt-a-spec/

vinyfalcao:
Hello everyone, it was a domain expired problem, we already renew and update the DNS in the host provides, in the next hours should work again...
Sorry about it :(

ivargrimstad:
Works fine now. Thanks!

PR #1405, opened 2022-02-09
Updated JUG PE website

linkfang:
For https://github.com/jakartaee/jakarta.ee/issues/1401


ivargrimstad:
It was a problem on their side. Has been fixed now, so the original URL should be used.

PR #1421, opened 2022-02-28
Add videos and filter to videos page

### LINKED ISSUE #1161: Video resource filter dropdown
On the video resources page there is a new feature that filters visible videos by a small JS widget that allows videos to be actively filtered by categories. This will be a new JS functionality that allows us to display and hide videos based on category. While we have similar search function filters in other sites, they have additional features that aren't currently needed, and it would likely be easier to write a short jq function to target and filter a set of containers.

An option for implementation is to add a `data-target` to the dropdown that indicates the container to be filtered, and a `data-filter-attr` with the values that can be filtered. Right now we can simply just hide and show the videos, but we may want to consider using short slide-in animations after hiding all videos. This will create a nicer effect of videos not just popping in and hide any processing time we may have.
 
Design: 
design

ericpoirier:


ericpoirier:
We have +1 from Shabnam.

chrisguindon (approved):

PR #1453, opened 2022-04-07
** DO NOT MERGE** Spec repo guidelines

ivargrimstad:
This is a work **in progress** for defining a guide for how to manage GitHub repositories under the Jakarta EE Working Group.

### DIFF
--- a/content/committees/specification/repositories/index.md
+++ b/content/committees/specification/repositories/index.md
@@ -11,7 +11,7 @@ aliases:
 
 - [Introduction]({{< ref "#intro" >}})
 - [GitHub Organizations]({{< ref "#gh-org" >}})
-- [Naming]({{< ref "#naming" >}})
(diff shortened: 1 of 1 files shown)


ivargrimstad:
> I do not see any need to align these names, actually. At JAX-RS we live perfectly well with having our own names and having TCK and API in the same repo. Moreover, we do use that for adding new features _in a single commit_ to both, API and TCK. So please do not standardize things that do not have an urgent need to be standardized. Thanks.

Thanks for your input, @mkarg!
There will not be a requirement regarding the naming of the repositories, not even how they're organized. If a project wishes to name it differently, and have the TCK and API in the same repository, have at it. 
These guidelines are just suggestions and conventions that have proved to work for most. They are here to help those that doesn't have strong opinions, or want to have some guidance.

PR #1456, opened 2022-04-12
Update featured stories

### LINKED ISSUE #1451: Add images to the featured story section of the Jakarta EE homepage via newsroom
Currently we are not able to add images in the newsroom to be featured on the homepage's featured story section. Adding images to that section can make the stories more visible to have a greater impact.

ericpoirier:
Fixes #1451 

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -44,9 +44,12 @@ document.addEventListener("DOMContentLoaded", function(event) {
         }
       });
 
+      var owlSolsticeSlider = $('.solstice-slider');
+      owlSolsticeSlider.owlCarousel();
+
       const solsticeSliderHome = () => {
-        var owl = $('.solstice-slider');
-        owl.owlCarousel({
+        var owlSolsticeFeaturedStorySlider = $('.solstice-featured-story-slider');
+        owlSolsticeFeaturedStorySlider.owlCarousel({
             items:1,
             autoplay:true,
             autoplayTimeout:6000,
@@ -58,13 +61,6 @@ document.addEventListener("DOMContentLoaded", function(event) {
       
       $(window).on("load", function() {
           solsticeSliderHome();
-      });
-      
-      $("body").on("shown.ef.featured_story", function(e) {
-          var owl = $('.solstice-slider');
-          owl.trigger('destroy.owl.carousel');
-          solsticeSliderHome();
-          owl.trigger('refresh.owl.carousel');
       });
 
       $(document).ready(function() {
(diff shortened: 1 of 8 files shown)


ericpoirier:
We have the +1 from @shabnammayel

js/main.js
```diff
@@ -44,14 +44,27 @@ document.addEventListener("DOMContentLoaded", function(event) {
...
+            autoplaySpeed: 2000,
+            loop:true,
+        });
+      }
+      
+      $(window).on("load", function() {
+          solsticeSliderHome();
+      });
+      
+      $("body").on("shown.ef.featured_story", function(e) {
```
chrisguindon: What is the purpose of this? 

I feel this we are doing something wrong when I see owl.trigger('destroy.owl.carousel'); then  owl.trigger('refresh.owl.carousel');

ericpoirier: its to make sure the carousel is loaded after the content is added. This is also something we are doing on ECDtools

chrisguindon: I feel this is the way we should do it:
https://www.jq22.com/demo/OwlCarousel2/demos/ajax.html

There is an event that you can trigger for inserting content.

ericpoirier: I'll give it a try.

ericpoirier: turns out I was using the same class for 2 different carousel on the page which was causing conflicts.

I update the class name for the Featured Story slider and removed the unnecessary code and both sliders are working well.

js/main.js
```diff
@@ -44,14 +44,23 @@ document.addEventListener("DOMContentLoaded", function(event) {
...
-          autoplay:true,
-          autoplayTimeout:6000,
-          autoplayHoverPause:true,
-          autoplaySpeed: 2000,
-          loop:true,
+      var owlSolsticeSlider = $('.solstice-slider');
+      owlSolsticeSlider.owlCarousel();
+
+      const solsticeSliderHome = () => {
+        var owlSolsticeFeaturedStorySlider = $('.solstice-featured-story-slider');
```
chrisguindon, line 51: We don't/shouldn't be using var here.

I still think that the solution that I provided makes more sense but this is better than the previous revision.

$(window).on("load", function() { is fired when the whole page has loaded, including all dependent resources such as stylesheets and images.
 
 
 The solution that I provided can insert the owl carrousel once we receive the data from the API.

ericpoirier: I tried using the solution you suggested and it didn't seem work. The content is being added to the page but Owl isn't updated to create the carousel.

chrisguindon: You are right, this looks like an outdated version of the docs.

chrisguindon (approved):

PR #1497, opened 2022-06-28
Correct the facelettaglibrary to web-facelettaglibrary_4_0.xsd

starksm64:


pzygielo:
For me this shows:
- source repo is missing web-facelettaglibrary_4_0.xsd, perhaps its rename from 3.0 to 4.0 was missed in https://github.com/eclipse-ee4j/jakartaee-schemas/pull/15
IMO this should be fixed in upstream first.
- this repo - currently published [web-facelettaglibrary](https://jakarta.ee/xml/ns/jakartaee/web-facelettaglibrary_3_0.xsd) for [JakartaEE 9](https://jakarta.ee/xml/ns/jakartaee/#9) references `<xsd:include schemaLocation="jakartaee_10.xsd"/>` which does not seem to be right and should be reverted to 9. Just to clarify - all changes to w-ftl-3.xsd should be reverted, not the single `include` only.

Similar to above - missing in jakartaee-schemas, JakartaEE 9 schema including Jakarta EE 10 - for https://jakarta.ee/xml/ns/jakartaee/web-facesconfig_4_0.xsd and https://jakarta.ee/xml/ns/jakartaee/web-facesconfig_3_0.xsd accordingly.

starksm64:
I fixed the ee10 ref to ee9 in the 3_0 facelettaglibrary, but I don't understand the point of the jakartaee-schemas when it has only ever contained a fraction of the schemas, and cannot be the true source since the specification projects are what generate the schemas. I'll bring this up on tomorrows platform call.

pzygielo:
> I fixed the ee10 ref to ee9 in the 3_0 facelettaglibrary,

The commit https://github.com/jakartaee/jakarta.ee/pull/1497/commits/ccab423913037f39d537fc9b9fb7aa654680cc60 brings some changes, but that's not what I meant:
>> Just to clarify - all changes to w-ftl-3.xsd should be reverted, not the single include only.


> but I don't understand the point of the jakartaee-schemas when it has only ever contained a fraction of the schemas, and cannot be the true source since the specification projects are what generate the schemas.

I see. So what is the source of file committed here then? I can't locate it in the tree https://github.com/jakartaee/faces/tree/4.0.1-RELEASE. It's not a part of https://jakarta.oss.sonatype.org/content/repositories/staging/jakarta/faces/jakarta.faces-api/4.0.1 artifact. But I might be wrong assuming it's Faces responsinble for that schema.

When the spec generates the schema - what happens next, where does it go? I also wonder what's the point of jakartaee-schemas if it's not to be used for that.

yersan:
@starksm64 What about `jakartaee_web_services_2_0.xsd`?

This XSD version is used for Jakarta EE 9 and Jakarta EE 10. However, it was updated to use the Jakarta EE 10 common types.

Would it be safe from the point of view of Jakarta EE 9 descriptor validations? They will be using the Jakarta EE 10 common types instead of the Jakarta EE 9.

pzygielo:
:confused:

PR #1506, opened 2022-08-03
replace http://www.w3.org/2001/xml.xsd with http://www.w3.org/2001/xm…

### LINKED ISSUE #1505: Redirecting to https on all of www.w3.org
**Is your feature request related to a problem? Please describe.**

Tests of [Redirecting to https on all of www.w3.org](https://www.w3.org/blog/2022/07/redirecting-to-https-on-www-w3-org/) today caused some problems with DD validation using `xml-maven-plugin`:
- https://github.com/eclipse-ee4j/openmq/pull/1549#issuecomment-1201574549

due to
- https://github.com/jakartaee/jakarta.ee/blob/8981afdf02d15050b1f51e4d3eb9b8ce01654a3f/static/xml/ns/jakartaee/jakartaee_10.xsd#L52 using http
- and https://github.com/mojohaus/xml-maven-plugin/issues/59.

**Describe the solution you'd like**
I'd like jakartaee xsds to use https for w3 locations instead.

**Describe alternatives you've considered**

**Additional context**

ivargrimstad:
Will close #1505

### DIFF
--- a/static/xml/ns/jakartaee/jakartaee_10.xsd
+++ b/static/xml/ns/jakartaee/jakartaee_10.xsd
@@ -49,7 +49,7 @@
   </xsd:annotation>
 
   <xsd:import namespace="http://www.w3.org/XML/1998/namespace"
-              schemaLocation="http://www.w3.org/2001/xml.xsd"/>
+              schemaLocation="https://www.w3.org/2001/xml.xsd"/>
 
   <xsd:include schemaLocation="jakartaee_web_services_client_2_0.xsd"/>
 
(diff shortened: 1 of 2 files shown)


pzygielo:
:+1: That would solve my case indeed. Thanks!

However 
1. on the W3 blog some comments were posted, by other parties observing similar issue. I'm not sure how W3 will proceed with their plan, BUT the location WITH https IS correct so this change looks fine for me, even if they decide to keep http for another decade
1. JakartaEE9 - I don't know what's the process of changing **published** schema, given the redirect was turned on only for the test, for 8 hours, and now http works fine again

ivargrimstad:
If http works fine for EE9, maybe we should leave that schema as it is and change it if someone complains in the future?

pzygielo:
> If http works fine for EE9, maybe we should leave that schema as it is and change it if someone complains in the future?

Fully agree.

PR #1606, opened 2022-12-20
Update configs to use split files for better mgmt

autumnfound:
Migrated to use broken down files as site is getting complicated and harder to manage. This makes each section far easier to manage and easier to extend. This is done in anticipation for the next language applied to the site.


oliviergoulet5:
Seems like a good idea. LGTM

ericpoirier (approved):

autumnfound:
Will need to be rechecked as I had to rebase but we should be clear as the only changes were menu changes for Why Jakarta EE and the resources section changes

ericpoirier (approved):

PR #1622, opened 2023-02-02
Add news about DEVIES award to Jakarta EE 10

ivargrimstad:


ivargrimstad:
Seems we're publishing news in another way these days...

PR #1647, opened 2023-03-24
Hotfix button group buttons not having equal height

### LINKED ISSUE #1588: Bug: Buttons are not equal height
Navigate to https://jakarta.ee/projects/

The button group's dropdown button is expected to have the same height as the download button. Currently it looks like this:
image

oliviergoulet5:
Resolves #1588 

This is fixed in EFHST 0.0.183 but it also includes Hugo v110 and Node 18. We're not going to upgrade the Hugo and Node version here just yet.

In the meantime, I suggest we add this hotfix.

### DIFF
--- a/less/styles.less
+++ b/less/styles.less
@@ -295,3 +295,14 @@ div.card-panel.panel div.panel-body.panel-body-with-img {
   font-weight: 700;
   margin-top: 10px;
 }
+
+// Hotfixes
+.project-summary {
+  .btn-sm {
+    .btn;
+    
+    &:hover {
+      color: @white;
+    }
+  }
+}


oliviergoulet5:
This affects the projects page: https://deploy-preview-1647--jakartaee.netlify.app/projects/

less/styles.less
```diff
@@ -294,4 +294,15 @@ div.card-panel.panel div.panel-body.panel-body-with-img {
   color: #F98200;
   font-weight: 700;
   margin-top: 10px;
+}
+
+// Hotfixes
```
chrisguindon, line 299: I feel this is a hack. I think we simply need to remove the btn-sm class on 

`<button type="button" class="btn btn-sm btn-warning dropdown-toggle details-link" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">`
            <span class="caret"></span>
            <span class="sr-only">Toggle dropdown</span>
          </button>

oliviergoulet5: We can't modify the class list because the shortcode is defined on EFHST. What you brought up, I did in a [patch on EFHST](https://gitlab.eclipse.org/eclipsefdn/it/webdev/hugo-solstice-theme/-/commit/bcdbbb45d95bf3081b80cecdd5a72b1855d732cf) but that would require an upgrade to the Hugo version. I was thinking this could be a temp fix until we are ready to do that.

chrisguindon: I am fine with a temp solution as long as you create an issue to clean up once we upgrade the hugo theme!

oliviergoulet5: Here is the issue to keep track of the cleanup: https://github.com/jakartaee/jakarta.ee/issues/1649

PR #1673, opened 2023-05-05
Update logos to include registered trademark

### LINKED ISSUE #1672: Update Jakarta EE logo with registered trademark versions
The header logo needs to display the registered trademark logo as opposed to the unregistered version. Same thing for the "share image" for social media and the Jakarta EE Compatible logo.

oliviergoulet5:
Resolves #1672


chrisguindon:
I am fine with switching to .svg but some webpages (internal and/or external) might be using these logo files that you are deleting. I am thinking it would be safer to update them but replace them with the latest version.

oliviergoulet5:
> I am fine with switching to .svg but some webpages (internal and/or external) might be using these logo files that you are deleting. I am thinking it would be safer to update them but replace them with the latest version.

I was afraid that would be the case... My goal was to clean this directory up with the logo name convention that Mehrnaz uses. 

I'll create another patch which just updates the images instead of deleting.

PR #1686, opened 2023-06-19
Add support for page fragments to Jakarta site

autumnfound:
Using new version of Hugo theme, a new templates section was added that allows the sites base page template to be retrieved programatically. These pages do not contain any page content, and will need to be added manually to be useful.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,13 +26,15 @@
     "optimize_news_images": "bash ./optimize_images.sh content/news"
   },
   "dependencies": {
-    "eclipsefdn-hugo-solstice-theme": "0.0.186",
+    "eclipsefdn-hugo-solstice-theme": "^0.0.187",
     "list.js": "^1.2.0",
-    "lodash-es": "^4.17.21",
-    "querystring-es3": "^0.2.1"
+    "lodash-es": "^4.17.21"
   },
   "devDependencies": {
     "vue-template-compiler": "2.6.12"
+  },
+  "overrides": {
+    "semver": "^7.5.3"
   },
   "browserslist": "last 5 version, > 0.2%, not dead, IE 11"
 }
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "optimize_news_images": "bash ./optimize_images.sh content/news"
   },
   "dependencies": {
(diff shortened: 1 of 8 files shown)


autumnfound:
This is updated to support the template pages changes I made in the base theme. I don't see any issues on a cursory glance through the site. Was there a specific feature you were worried might not work in the upgrade?

package.json
```diff
@@ -26,7 +26,7 @@
     "optimize_news_images": "bash ./optimize_images.sh content/news"
   },
   "dependencies": {
-    "eclipsefdn-hugo-solstice-theme": "0.0.181",
+    "eclipsefdn-hugo-solstice-theme": "0.0.186",
```
chrisguindon: We should also update the Hugo version in the netlify file if we upgrade this package. See https://gitlab.eclipse.org/eclipsefdn/infrazilla/-/issues/1011 for more info.

We also need to update the readme.md when we make this change:
https://gitlab.eclipse.org/eclipsefdn/it/websites/eclipse.org/-/commit/f85703d8ff207a74cdc4ca0a3bd2e02549d52951

Finally, we had some issues with the spec project:

TLDR; yarn is not installed for that project. Can you also switch form yarn.lock to packages-lock.json with this change?

webpack.mix.js
```diff
@@ -15,11 +15,20 @@ require('./node_modules/eclipsefdn-solstice-assets/webpack-solstice-assets.mix.j
 let mix = require('laravel-mix');
 mix.EclipseFdnSolsticeAssets();
 
+// required to resolve missing polyfills
```
chrisguindon: If this is going to be needed for all the sites, let's fix it in solstice-assets so that all sites can benefit: 
https://gitlab.eclipse.org/eclipsefdn/it/webdev/solstice-assets/-/blob/main/webpack-solstice-assets.mix.js

chrisguindon (approved):

chrisguindon:
> This is updated to support the template pages changes I made in the base theme. I don't see any issues on a cursory glance through the site. Was there a specific feature you were worried might not work in the upgrade?

The specifications repo project. Once this is merged, we will need to upgrade the Hugo version and perhaps some build commands.

https://github.com/jakartaee/specifications/blob/master/netlify.toml

Issue #1717, opened 2023-09-12
Header Banner Missing


excentrickristy:
I noticed today that these two pages are missing the blue header banner:
https://jakarta.ee/about/jesp/
https://jakarta.ee/committees/specification/guide/

While that is being fixed, could we add a global bottom margin to the inner pages header banner? Maybe 30 or 40px?
I noticed that on many pages the content is getting very close to the banner:
<img width="1190" alt="image" src="https://github.com/jakartaee/jakarta.ee/assets/73264129/6b281cf3-9573-4345-a97b-07f959d0bcd9">

excentrickristy:
Also missing the header banner: 

https://jakarta.ee/committees/specification/compatibility/
https://jakarta.ee/committees/specification/operations/
https://jakarta.ee/membership/join-us/
https://jakarta.ee/committees/specification/
https://jakarta.ee/community/developer-advocate/
https://jakarta.ee/committees/steering/brand_use_by_projects/  - this page is also missing the banner above the footer:
<img width="1380" alt="image" src="https://github.com/jakartaee/jakarta.ee/assets/73264129/fb500c00-960a-4ea4-ab0a-dc943e3ad144">

oliviergoulet5:
Thanks for the list of pages @excentrickristy.

Here is a preview of the header banners being added to the pages you listed:

- https://deploy-preview-1721--jakartaee.netlify.app/about/jesp/
- https://deploy-preview-1721--jakartaee.netlify.app/committees/specification/guide/
- https://deploy-preview-1721--jakartaee.netlify.app/committees/specification/compatibility/
- https://deploy-preview-1721--jakartaee.netlify.app/committees/specification/operations/
- https://deploy-preview-1721--jakartaee.netlify.app/committees/steering/brand_use_by_projects/
- https://deploy-preview-1721--jakartaee.netlify.app/community/developer-advocate/
- https://deploy-preview-1721--jakartaee.netlify.app/membership/join-us/

The margin problem with the header banner occurred from a bug with missing breadcrumbs. I merged a patch which fixed the margins and the missing breadcrumbs.

However I could not reproduce the featured footer (the component with the "Register for EclipseCon 2023" text) problem you had. The background is an image so it is possible it wasn't loaded due to some network problem at the time. If it is still occurring, try to clear your cache with a hard refresh.

excentrickristy:
[...]

PR #1722, opened 2023-09-18
fix: Breadcrumbs missing

oliviergoulet5:
Relates to #1717 

Breadcrumbs were missing from the site. This is because the breadcrumbs were using Astro's layout for some reason. Luckily, setting `layout_style` to `quicksilver` will fix this.

Ultimately, this seems to be a bug in EFHST. We shouldn't be using the Astro breadcrumbs by default.

### DIFF
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -10,6 +10,7 @@ enableRobotsTXT = true
 enableInlineShortcodes = true
 
 [Params]
+  layout_style = "quicksilver"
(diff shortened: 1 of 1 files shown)


chrisguindon (approved):

chrisguindon:
> Ultimately, this seems to be a bug in EFHST. We shouldn't be using the Astro breadcrumbs by default.

+1 default should be quicksilver

PR #1727, opened 2023-09-29
fix: Prevent last nav menu from overflowing

### LINKED ISSUE #1726: Resources menu overflowing off the page at smaller screen widths
**Describe the bug**

The resources menu is overflowing off the page at smaller screen widths.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to https://jakarta.ee/
2. Resize window to 1280px width
3. Click "resources" in the navbar

**Expected behaviour**

It should not be overflowing. A solution could be for the menu to open toward the left instead of the right. 

**Screenshots**

image

oliviergoulet5:
Resolves #1726

### DIFF
--- a/less/layout/_navbar.less
+++ b/less/layout/_navbar.less
@@ -1,3 +1,17 @@
+/*!
+ * Copyright (c) 2019, 2023 Eclipse Foundation, Inc.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * Contributors:
+ *   Eric Poirier <eric.poirier@eclipse-foundation.org>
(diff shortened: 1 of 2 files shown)


less/layout/_navbar.less
```diff
@@ -213,6 +227,18 @@
     }
 }
 
+// Prevent menu from overflowing on the right side of the screen. The reason we
+// are selecting the 2nd last child is because the last child is a mobile-only
+// menu item. 
+#main-menu .navbar-nav li.dropdown:nth-last-child(2) {
```
oliviergoulet5, line 233: It might seem a bit hacky. I wanted to add a custom parameter to the menu item in menu.toml. Something like `class` or `direction`. However, Hugo doesn't allow params on menu items, only menus.

I also tried doing `:last-of-type`, but that selector only works with elements, not classes. That would've been perfect if it worked because I would select the last `.dropdown`.

Bootstrap introduced responsive menu direction classes in v4. When we migrate to v5, we could have a better solution using that.

PR #1761, opened 2023-11-16
japanese contents for homepage

kazumura:
This is a PR for only homepage.
A bunch of PRs will follow for the rest of contents.

### DIFF
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -58,7 +58,7 @@
         {{ if and (eq $has_cfa 1) (not (or (eq .Site.Params.hide_cfa_same_page "true") (eq $is_cfa_page 1) (eq $hide_call_for_action true))) }}
           <div class="{{ $.Site.Params.call_for_action_wrapper_classes | default "float-right hidden-xs" }}" id="btn-call-for-action">
             <div class="btn-starter">
-              <a class="btn btn-secondary margin-0" href="https://start.jakarta.ee/">{{ .Site.Params.call_for_starter_action_text }}</a>
+              <a class="btn btn-secondary margin-0" href="https://start.jakarta.ee/">Starter</a>
             </div>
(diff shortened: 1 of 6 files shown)


layouts/partials/navbar.html
```diff
@@ -58,7 +58,7 @@
         {{ if and (eq $has_cfa 1) (not (or (eq .Site.Params.hide_cfa_same_page "true") (eq $is_cfa_page 1) (eq $hide_call_for_action true))) }}
           <div class="{{ $.Site.Params.call_for_action_wrapper_classes | default "float-right hidden-xs" }}" id="btn-call-for-action">
             <div class="btn-starter">
-              <a class="btn btn-secondary margin-0" href="https://start.jakarta.ee/">Starter</a>
+              <a class="btn btn-secondary margin-0" href="https://start.jakarta.ee/">{{ .Site.Params.call_for_starter_action_text }}</a>
```
chrisguindon: @kazumura I see that you are copying the solution from the call_for_action_text. This is code that was written before we started to translate the site.

The best practice in Hugo is to translate a string using i18n: https://gohugo.io/content-management/multilingual/

This would require us to create a new file in the https://github.com/jakartaee/jakarta.ee/tree/src/i18n folder for this new Japanese translation. This is needed to fully translate the site. A good way to get started is by copying the English file to create the jp.toml file.

When creating a new translatable string, we also need to make sure that we add an entry in the default language which is English to make sure the English website is not missing any text.

While your solution is not completely wrong since it's used elsewhere. I felt it would be a good idea to raise the best practice early as I know you plan to make many translations in the short term.

chrisguindon:
While revising this PR, I noticed that the language switcher is not working properly on the site. I created the following issue to fix that: https://github.com/jakartaee/jakarta.ee/issues/1762

layouts/partials/navbar.html
kazumura: @chrisguindon 
I have several experiences about multi language in Hugo with JakartaOne Livestream website, and
I know the file tree structure and i18n folder in Hugo.
This PR is only minmum changes as the first step.
We will create i18n/ja.toml after this PR.

chrisguindon: @kazumura My feedback is quite simple with a goal of following best practices.

I would appreciate it if you would be open to accepting feedback so that we can collaborate!

kazumura: @chrisguindon 
I give up "Starter" for now.

PR #1784, opened 2023-12-08
[ja] Add compatible products pages

kaido207:
Add compatibility pages.
And modify that the pages link `get-listed` was wrong for all lang.


kazumura (approved):
thanks

PR #1785, opened 2023-12-08
[ja] Translate pages under /committees/specification into Japanese

tnagao7:
This PR translates contents under `/committees/specification` into Japanese.

Since translation is based on contents as of several months ago, the following pages have not yet been fully translated:

- `/committees/specification/guide/`:  a part of the "Introduction" section has not been translated yet
- `/committees/specification/tckprocess/`": not version 1.4 but 1.3 of Jakarta EE TCK Process is translated
- `/committees/specification/versioning/`: not yet translated


kazumura (approved):
thanks

Issue #1820, opened 2024-01-11
Add the Jakarta RESTful Web Services Video to the spec page


shabnammayel:
Please add the new video about the Jakarta RESTful web service to the spec page on the RESTful Web service box, similar to what we did with the Platform video: https://www.youtube.com/watch?v=7mfVyPI360k

Please add a box next to the website box on the Jakarta RESTful Web service section and call it "Latest Updates" and add the link to the video there: https://jakarta.ee/specifications/ 

We will soon create an issue to add all these learning videos on the new "Learn" Page as well. thanks

shabnammayel:
@oliviergoulet5 and @chrisguindon can we please move forward with this? thanks

oliviergoulet5:
The video guide went live on the site yesterday: https://jakarta.ee/specifications/

PR #1822, opened 2024-01-17
feat: add latest video button to specifications

oliviergoulet5:
Resolves part of #1820

### DIFF
--- a/less/components/_buttons.less
+++ b/less/components/_buttons.less
@@ -11,19 +11,6 @@
  * SPDX-License-Identifier: EPL-2.0
 */
 
-//== Buttons
-.btn {
-  padding: 0.8rem 1.5rem;
-}
-
-.btn-huge {
-  padding: 1rem 3.5rem;
-}
-
-.btn-xs {
-  padding: 0.1rem 0.5rem;
-}
-
 //== Contained button variant
 .btn-contained {
   border: 1px solid #fff;
--- a/less/components/_buttons.less
+++ b/less/components/_buttons.less
@@ -20,6 +20,10 @@
   padding: 1rem 3.5rem;
 }
 
+.btn-xs {
+  padding: 0.1rem 0.5rem;
+}
+
 //== Contained button variant
 .btn-contained {
   border: 1px solid #fff;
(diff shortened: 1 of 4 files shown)


chrisguindon (approved):

oliviergoulet5:
This patch adds a button labelled "Latest Updates" if a specification page has the `latest_updates_video` front matter param.

image

It also fixes a bug where `.btn-xs` was not smaller than a regular `.btn`.

less/components/_buttons.less
```diff
@@ -20,6 +20,10 @@
   padding: 1rem 3.5rem;
 }
 
+.btn-xs {
+  padding: 0.1rem 0.5rem;
```
chrisguindon: While you are fixing .btn-xs, I suspect that the other buttons size are not working as expected because of 

.btn {
  padding: 0.8rem 1.5rem;
}
The bootstrap project utilizes variables and the .button-size mixing to set the padding
https://github.com/twbs/bootstrap/blob/v3-dev/less/buttons.less

oliviergoulet5: I didn't want to touch those because other components could also use `@padding-base-vertical` and the other variables. There isn't a variable dedicated to buttons in BS3.

chrisguindon: You should be able to search for the less code and confirm that.  My main concern here is that we are overriding the padding with .btn. which is overriding all the .btn size classes to use the same padding. I suspect that .btn-lg and .btn-sm  are not working as expected anymore.

oliviergoulet5: Added an MR for EFSA to add dedicated button variables: https://gitlab.eclipse.org/eclipsefdn/it/webdev/solstice-assets/-/merge_requests/473

ivargrimstad:
@TanjaObradovic  @shabnammayel  Do we really want to call it "Latest Updates"?

In a couple of months, a new version will be available, but the video will still be about older content, or do we expect to always have a current video? 

Wouldn't be better to simply call it "Watch Video" or something?

shabnammayel:
> @TanjaObradovic @shabnammayel Do we really want to call it "Latest Updates"?
> 
> In a couple of months, a new version will be available, but the video will still be about older content, or do we expect to always have a current video?
> 
> Wouldn't be better to simply call it "Watch Video" or something?

Agreed! Let's change the name to Video Guide everywhere @oliviergoulet5 thanks

ivargrimstad (approved):

Issue #1813, opened 2024-01-04
Update TCK Process to specify that Jakarta EE Platform implementations only need to pass TCK tests that can be run against a Platform (e.g. EE containers) implementation to be considered Jakarta EE Platform compatible. 


scottmarlow:
**Describe the bug**
As per https://github.com/jakartaee/jakarta.ee/pull/1679#issuecomment-1728181551 feedback, I think the (Jakarta EE) TCK Process requirements should be updated to not require Standalone SE TCKs to include Platform tests:
```
Jakarta EE Platform implementations only need to pass TCK tests that can be run against a Platform (e.g. EE containers) implementation to be considered Jakarta EE Platform compatible. Jakarta EE Platform implementations do not need to run TCK tests that do not run against a Platform implementation.
```

**Expected behavior**

The current Jakarta EE development process involves completing (included) Specification releases in a series of waves that are respectful of the needed compilation dependencies.  

If we follow the new TCK Process changes with the [EE 11 development process](https://jakartaee.github.io/platform/jakartaee11/JakartaEE11ReleasePlan), each Specification wave TCK needs to include testing on a EE 11 Platform implementation which cannot happen as waves 1-6 will definitely not have an EE 11 Platform implementation to test against.

I recommend that we either correct the TCK Process or revert the https://github.com/jakartaee/jakarta.ee/pull/1679 change so that Standalone Specifications can proceed through the ballot process without having an EE 11 Platform implementation to test against.  

Correcting the TCK Process is not difficult, the hard thing for me is that my feedback seemed to be ignored on the https://github.com/jakartaee/jakarta.ee/pull/1679 which perhaps is due to me not being clear enough with my feedback.

Emily-Jiang:
@scottmarlow I don't quite understand your concern here. The updated process is to prevent from only SE style tcks available per spec. It does not say that all TCKs must pass to release a spec. The integration tests are normally in platform tck suite and it will run at the later stage. There is nothing new here. Take CDI as an example, it has integration tests with EJB etc. As long as that tests pass, the spec can be released.

scottmarlow:
> @scottmarlow I don't quite understand your concern here. The updated process is to prevent from only SE style TCKs available per spec. 

A specific example is the https://download.eclipse.org/jakartaee/persistence/3.1/jakarta-persistence-tck-3.1.0.zip SE style TCK which as per the 1.4 TCK Process change is now required to be changed to run against `standalone or in a platform (or profile).`

[...]

PR #1852, opened 2024-02-13
issues/1813 Update TCK Process to unblock ballot release of EE 11 Specification ballots that have traditionally been done with Java SE style TCKs.

scottmarlow:
See discussion on https://github.com/jakartaee/jakarta.ee/issues/1813.

If this change is made or a change like it, we should follow up to make a further change to cover `Update TCK Process to specify that Jakarta EE Platform implementations only need to pass TCK tests that can be run against a Platform (e.g. EE containers) implementation to be considered Jakarta EE Platform compatible.`  This change might not really belong in the TCK Process guide as it is specific to Platform/Profile TCKs.  I suggest that the https://github.com/jakartaee/platform-tck/tree/tckrefactor might be a better place to specify the requirement which does imply more work will be done for the Platform TCK to support running such tests (e.g. Core Profile 10 TCK already has an example of doing that).

CC @Emily-Jiang @lukasj

### DIFF
--- a/content/committees/specification/tckprocess/index.md
+++ b/content/committees/specification/tckprocess/index.md
@@ -51,11 +51,7 @@ There are 3 different types of TCKs, detailed below.
 - Type 2: Tests that only apply when running in a platform (or profile) to test behavior defined by the specification for that mode of execution.
 - Type 3: Tests that apply to the behavior defined by a specification for the execution either as a standalone or in a platform (or profile).
 
-The individual specification TCK must contain Type 2 or Type 3 tests while Type 1 tests are optional. A profile-ready mechanism must be provided for running the required TCK tests (types 2 and 3) in the profiles they are included in: platform, web profile or core profile. Type 1 tests must be excluded and not required to be executed.
-
-The individual specification TCK must provide a mechanism for running the required TCK tests (types 1 and 3) outside the context of a platform or profile. Type 2 tests must be excluded and not run.
-
-The type 3 tests must be executed in compatible implementations of the profiles to validate the individual specification implementation functions properly in the profile implementation: platform, web profile or core profile.
+A profile-ready mechanism must be provided for running the required TCK tests in the profiles they are included in: platform, web profile or core profile.
 
 ## Materials for a TCK Release {#_materials_for_a_tck_release}
 


Emily-Jiang (changes requested):
I disagree with this change. The reason to add the statement is to prevent from specs just having standalone tcks. I would rather the spec fixing the issues rather than pushing the releases out.

bstansberry:
@scottmarlow @Emily-Jiang @lukasj @edburns 

I think it would be good to have a discussion going through the implications of the existing text on the various specs as a way of better understanding what's meant by the text and how it will affect EE 11 and service releases.

I believe I understand what's intended by the existing text, but working through some concrete cases will give me more confidence that I really do. As a person primarily concerned with implementing the profiles and full platform I certainly appreciate the general thrust of wanting to ensure that relevant tests can run in an EE environment, particularly as we continue to work to move spec-specific tests out of the platform TCK.

I have no sense at all though of how much effort and time getting the various TCKs to comply with this will require, and since EE 11 is an phase where deadlines are imminent I think it would be good if we had more clarity around that.

lukasj:
XML-Binding can be used as an example affected by current wording should it want to produce an update release. The TCK for this particular spec is and always has been SE based, unless I'm mistaken, and any, even simple, change there requires non-trivial effort. Simply re-running all tests within all possible existing _vehicles_ (or profile/container combination if you wish) would lead to an insane number of hours required to validate vendor's EE runtime (likely more than a day) - I do not think that is a right way to go. Alternatively it could be said that XML-WS TCK is used to validate XML-B runtime, since it XML-WS spec is heavy user of XML-B, but XML-WS is just an optional part of the platform, so not every vendor has an option to use it.

As for implications on current service releases - given the service release is based on EE 10 artifacts, updated rules should not be applicable as the change in the process should not be retroactive.

bstansberry:
> As for implications on current service releases - given the service release is based on EE 10 artifacts, updated rules should not be applicable as the change in the process should not be retroactive.

Makes sense to me. TBH I'm not an EE process expert, so maybe this is already a clear rule. If not we should make it one, and if it is already a rule it's good to socialize it a bit to be sure others are aware.

XML Binding seems like the extreme case in terms of the implications of these requirements, so extreme that I can imagine just in the interest of expediency it ends up getting a permanent exemption. :) I'm just one voice on that though, and not a voting voice on any relevant committee. If service releases are not affected it gets at least a temporary exemption, as there's no major/minor release planned for EE 11.

But I'm sure there are other specs doing major/minor releases for EE 11 that will be affected.

Emily-Jiang:
> > As for implications on current service releases - given the service release is based on EE 10 artifacts, updated rules should not be applicable as the change in the process should not be retroactive.
> 
Yes, any service release is not affected by the new process.
> Makes sense to me. TBH I'm not an EE process expert, so maybe this is already a clear rule. If not we should make it one, and if it is already a rule it's good to socialize it a bit to be sure others are aware.
> 
> XML Binding seems like the extreme case in terms of the implications of these requirements, so extreme that I can imagine just in the interest of expediency it ends up getting a permanent exemption. :) I'm just one voice on that though, and not a voting voice on any relevant committee. If service releases are not affected it gets at least a temporary exemption, as there's no major/minor release planned for EE 11.
> 
> But I'm sure there are other specs doing major/minor releases for EE 11 that will be affected.

This process will be in place for any TCK rework and new TCKs. For the XML Binding spec, we can look into the details and might need to make it exception. As a general guidance, we should design TCKs to be executed by Jakarta EE runtimes.

PR #1861, opened 2024-02-27
feat: update videos page

### LINKED ISSUE #1731: Modify videos page to read YouTube RSS Feed
On this page: https://jakarta.ee/community/videos/

Our initial idea for this page is that it would update automatically when YouTube videos are posted. If I'm reading the code right, it looks like updating the page is a manual process at the moment.

Can we modify the video page to update automatically by reading the YouTube XML feed for each playlist?

Here is the XML feed for the Jakarta Tech Talk Playlist: https://www.youtube.com/feeds/videos.xml?playlist_id=PLutlXcN4EAwC64sgFLJSWAgQJvVo6T4Dh

We could associate a feed with each playlist we want to display in the dropdown.

oliviergoulet5:
Resolves #1731 

Fetches latest videos from playlists. This could not be implemented in Javascript because of CORS so I made a small script to have it fetch the data at build time.


fetch_videos.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
```
chrisguindon, line 2: Make sure to include copyright headers in this file but other than that this looks good.

oliviergoulet5: Done

fetch_videos.sh
```diff
@@ -0,0 +1,26 @@
...
+  "PLutlXcN4EAwDCVHTlhymKrKBwSyddZoc8" # Update Calls 
+  "PLutlXcN4EAwAfZaPt8pdamYvE7APRcPbA" # Livestream 2024 PT
+  "PLutlXcN4EAwDNN2lVNlfV3u2rBvJMOFCn" # Livestream 2023 EN
+  "PLutlXcN4EAwD_rW8juyr1ZQf7MEYRQuWb" # Livestream 2023 ES
+  "PLutlXcN4EAwCGe7GN9htD_QvANU9xTvrh" # Livestream 2023 ZH
+)
+
+mkdir -p ./data/videos
+
+for playlist_id in "${playlist_ids[@]}"; do
```
chrisguindon, line 31: I was thinking about this last night.

We should handle errors in a way that if the curl command fails during a build, the CI build fails to avoid deploying a site without any videos.

We should be able to do that with exit 1:

```
# Loop through each playlist ID and download its XML feed
for playlist_id in "${playlist_ids[@]}"; do
  # Download XML feed for the playlist
  curl_output=$(curl -s "https://www.youtube.com/feeds/videos.xml?playlist_id=$playlist_id")
  
  # Check if curl command was successful
  if [ $? -eq 0 ]; then
    # Save XML feed to file
    echo "$curl_output" > "./data/videos/$playlist_id.xml"
    echo "Downloaded XML for playlist: $playlist_id"
  else
    echo "Failed to download XML for playlist: $playlist_id"
    exit 1  # Return non-zero exit code to indicate failure
  fi
done

```

oliviergoulet5: Good idea

oliviergoulet5: Added the error handling as well as additional comments.

.gitignore
```diff
@@ -11,6 +11,7 @@
 /layouts/robots.txt
 /.project
 /data/eclipsefdn_projects.yaml
+/data/videos/
```
chrisguindon, line 14: good catch! I wanted to mention that and I forgot :)

chrisguindon:
@oliviergoulet5 Great job coming up with this solution and implementing it so quickly!

chrisguindon (approved):

PR #1874, opened 2024-03-14
remove www. from cononical url

chrisguindon:


ZacharySabourin (approved):

chrisguindon:
www. is not the default

PR #1954, opened 2024-07-11
fix: event image overflow

oliviergoulet5:
Resolves https://gitlab.eclipse.org/eclipsefdn/it/webdev/solstice-assets/-/issues/419

### DIFF
--- a/js/templates/events-page.mustache
+++ b/js/templates/events-page.mustache
@@ -3,7 +3,7 @@
     <div class="row margin-bottom-50">
     {{#image}}
         <div class="col-sm-12">
-            <img class="img img-responsive" src="{{ image }}">
+            <img class="img-responsive" src="{{ image }}">
         </div>
     {{/image}}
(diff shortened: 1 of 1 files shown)


ZacharySabourin (approved):

js/templates/events-page.mustache
```diff
@@ -3,7 +3,7 @@
     <div class="row margin-bottom-50">
     {{#image}}
         <div class="col-sm-12">
-            <img src="{{ image }}">
+            <img class="img img-responsive" src="{{ image }}">
```
chrisguindon: Why adding .img? I don't think that class exist in our framework..

oliviergoulet5: I thought they were both needed. I've always been adding these two classes to images lol... but you're right. It does nothing.

I'll start removing them.

chrisguindon (approved):

PR #1976, opened 2024-08-22
fix: Ensure spec and docs are available on non-english spec pages

### LINKED ISSUE #1970: ZH Site - Broken Links in persistence 3.2
https://jakarta.ee/zh/specifications/persistence/3.2/ page points to non-existing pages:

```
Response 404: https://jakarta.ee/zh/projects/
Response 404: https://jakarta.ee/zh/specifications/persistence/3.2/apidocs
Response 404: https://jakarta.ee/zh/specifications/persistence/3.2/jakarta-persistence-spec-3.2.html
```

ZacharySabourin:
Resolves https://github.com/jakartaee/jakarta.ee/issues/1970

### DIFF
--- a/layouts/partials/head_custom.html
+++ b/layouts/partials/head_custom.html
@@ -19,12 +19,9 @@
 
 
 {{ if or (eq .Section "specifications") (strings.HasPrefix .Section "specifications/") }}
-  {{ if (strings.HasPrefix .RelPermalink "/zh/")}}
-    {{ with index (split .RelPermalink "/zh/")}}
-      <base href="/{{ index . 1 }}"/>
-    {{ end }}
-  {{ else if (strings.HasPrefix .RelPermalink "/ja/")}}
-    {{ with index (split .RelPermalink "/ja/")}}
+  {{ $lang_prefix := (print "/" .Language.Lang "/") }}
+  {{ if (strings.HasPrefix .RelPermalink $lang_prefix )}}
+    {{ with index (split .RelPermalink $lang_prefix)}}
       <base href="/{{ index . 1 }}"/>
     {{ end }}
   {{ end }}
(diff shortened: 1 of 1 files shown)


ZacharySabourin:
Thankfully the `.RelPermalink` method includes the language code at the beginning. With this I was able to catch non-English language codes and strip them from the spec documentation relative URLs.

For now, the only relative URLs on the spec pages point to documentation. With this fix, we will need to make sure this trend continues. Otherwise other URLs may be pointed to their English counterparts as well, which we might not want.

layouts/partials/head_custom.html
```diff
@@ -15,4 +15,17 @@
 
 <link rel="preconnect" href="https://fonts.googleapis.com">
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-<link href="https://fonts.googleapis.com/css2?family=Exo:wght@400;700&display=swap" rel="stylesheet">
\ No newline at end of file
+<link href="https://fonts.googleapis.com/css2?family=Exo:wght@400;700&display=swap" rel="stylesheet">
+
+
+{{ if or (eq .Section "specifications") (strings.HasPrefix .Section "specifications/") }}
+  {{ if (strings.HasPrefix .RelPermalink "/zh/")}}
```
autumnfound: Couldn't we just use the `.Language.lang` call from https://gohugo.io/methods/page/language/ to handle all current and future languages?

ZacharySabourin: Good call. I've updated it to be language agnostic

autumnfound:
Not sure how to validate the fix but the code LGTM

ZacharySabourin:
@autumnfound You can validate by visiting any spec page in a non-English language. E.g. https://deploy-preview-1976--jakartaee.netlify.app/zh/specifications/persistence/3.2/ and checking that the documentation links in the "Details" section function as expected. The goal is to point all spec pages to the English docs.

You can compare this against prod: https://jakarta.ee/zh/specifications/persistence/3.2/ where the same doc links yield a 404 outside of the English version.

While the 3.2 persistence docs were flagged as failing, it looks like this has always been a problem with non-English spec documentation. If you navigate to older specs, the problem is there as well.

autumnfound (approved):
Thanks for the steps, looks great!

PR #1986, opened 2024-09-10
fix build by removing scripts that fetch project data

chrisguindon:
@oliviergoulet5 - Netlify has lost access to our project API as of this morning. I implemented a patch to fix the build, but it results in the website badge being removed from the specification pages. 

Could you create a new web component that fetches the data client-side to allow us to drop the existing script that retrieves all the project data, while ensuring the website badge remains visible?

See an example here: https://jakarta.ee/specifications/webprofile/

Feel free to add commits to this MR to add the new component where I am removing `$.Site.Data.eclipsefdn_projects`

/cc @ivargrimstad


oliviergoulet5:
The website badge should now appear @chrisguindon

oliviergoulet5:
Also added a subtle error state if data could not be fetched for whatever reason:

image

layouts/specifications/list.html
```diff
@@ -12,23 +12,15 @@
     <h1 class="big-text">
       {{ .Title }}
       {{ if .Params.project_id }}
-        <span class="badge">
-          <a href="https://projects.eclipse.org/projects/{{ .Params.project_id }}"><i class="fa fa-archive" aria-hidden="true"></i> Project</a>
+        <span class="eclipsefdn-specification-badges" data-project-id="{{ .Params.project_id }}">
```
chrisguindon, line 15: Not sure why we are duplicating this 3 time in this file but I believe you need to add this widget in two more places before we can merge! Thanks for working on this!

oliviergoulet5: Np!

That's a good catch. I didn't notice it being used in multiple places. I added the possibility of having a video guide badge to the widget.

ZacharySabourin (approved):

PR #2069, opened 2024-12-18
fix: pagination not appearing

### LINKED ISSUE #1879: News page has no pagination
Here is Ivar's blog: https://blogs.eclipse.org/post/ivar-grimstad/jakarta-ee-developer-survey-2024

We are planning to add it to Jakarta EE's announcement section on the homepage. But we would like to add it to Jakarta's domain. Also, I just noticed this page only shows the recent items published https://jakarta.ee/news/ July 2023. Can we keep/display older news items? thanks

cc @jph152

oliviergoulet5:
Resolves #1879


oliviergoulet5:
There are a few regressions. I'll create another EFSA release which will iron these out.

PR #2127, opened 2025-03-28
Add GlassFish 8.0.0-M10 as compatible implementation of Jakarta EE 11 Web Profile

ivargrimstad:


anbusampath:
@ivargrimstad  can GlassFish-8.0.0-JDK17-M10 removed from [download](https://jakarta.ee/compatibility/download/) page since it is used for TCK only.

PR #2181, opened 2025-06-08
Set the correct fixed-value version attribute in the Jakarta Validation (3.1) XSDs

marko-bekhta:
Part of the fix for https://github.com/jakartaee/validation/issues/262


marko-bekhta:
thanks @ivargrimstad !

PR #2282, opened 2025-11-04
feat(mega-menu): adding base menu composition partial + styles

gnugomez:
In this PR I'm adding:

- Some base styles updates such as fonts & colors
- Updated navigation partial to properly render new mega menu
- Added some base styles for the mega menu

this isn't totally ready yet, and will experience changes.


config/_default/config.toml
```diff
@@ -24,20 +24,16 @@ enableInlineShortcodes = true
   favicon = "images/jakarta/favicon.ico"
   share_img = "images/jakarta/jakarta_ee_400x400.png"
   gcse = "011805775785170369411:p3ec0igo0qq"
-  subtheme = "jakarta"
```
ericpoirier, line 27: **Question:** You don't need the subtheme anymore?

gnugomez: I bet no, I think that isn’t actually being used right now; after removing it, I saw no change.

autumnfound: We're switching from quicksilver to astro, so there isn't even a concept for subthemes anymore. I do think this had knock on effects though, because the top of the release pages look pretty rough, e.g. https://deploy-preview-2282--jakartaee.netlify.app/release/10/

This is partially because the menu is different, but the general look and feel is fairly mangled on that page even outside of that

less/_variables.less
```diff
@@ -11,18 +11,26 @@
  * SPDX-License-Identifier: EPL-2.0
 */
 
-@brand-primary:         #1B208B; // Dark blue
-@brand-secondary:       #F98200; // Gold
-@brand-tertiary:        #58595B; // Dark Gray
-@brand-quaternary:      #FDB940; // Yellow 
+//=Primary Brand Colors
+@brand-primary:         #1B208B;
+@brand-primary-royal:   #182CB9;
```
autumnfound, line 16: This breaks our conventions, as we use the primary through quinary to denote the main colours, with each positional descriptor (i.e. primary, secondary) to indicate unique colours. While we do have some additional variants available, they are usually just darkened/lightened colours used for things like hover.

gnugomez: I’ll add the existing variables as polyfills, but I’ll use these names within the styles just to have a better semantics and to better align with the brand guide provided by the agency.

autumnfound (approved):

PR #2285, opened 2025-11-06
feat(mobile): adding full screen nested menu

gnugomez:


ericpoirier:
I haven't looked at the design but I feel like the text of the menu should align with the text of the logo.

image

gnugomez:
> I haven't looked at the design but I feel like the text of the menu should align with the text of the logo.
> image

I understand your concern, but the main issue with solving that is that this nav sticks to the scroll, and adds a white bg once you scroll a bit, now the text is aligned with the logo as a whole, but if I move ti down a bit it will then have to shift back to the center once we scroll, otherwise it might have too much white space on top, and the possible animation for that might be a bit odd, but I can try

image

This is the design with the white bg 

image


and this whit the blue/tranparent bg

image


although in the desing the nav text appears to be a bit closer to the logo text, maybe the logo is bigger, I'm not sure.

any suggestion on how to fix this @ericpoirier ?

layouts/partials/call_for_actions.html
```diff
@@ -0,0 +1,55 @@
...
+    {{ end }}
+    {{ $abs_cfa := printf "%s" ($normalized_cta_url | relURL) }}
+    {{ if or (eq $normalized_cta_url $.Permalink) (eq $abs_cfa $.Permalink) }}
+      {{ $is_cfa_page = true }}
+    {{ end }}
+  {{ end }}
+{{ end }}
+
+{{/* Render CTA buttons only if not hidden by page param or same-page logic */}}
+{{ if and .Site.Params.call_for_actions (not (or $hide_call_for_action $is_cfa_page)) }}
```
autumnfound: Now that CTAs are multiple values, we should consider that we may want to show the "Join us" CTA on the download page. Since we have the hide call to action param as well, we can just hide it on those pages if we really want none to show.

autumnfound (commented):
For new files, ensure copyright header is accurate. We should be using `Eclipse Foundation AISBL` now instead of Inc.

layouts/partials/nav_desktop.html
```diff
@@ -1,15 +1,13 @@
-<!-- 
-  Copyright (c) 2018 Eclipse Foundation, Inc.
+<!--
+ Copyright (c) 2025 Eclipse Foundation AISBL
```
ericpoirier: I think this should be 2018.

layouts/partials/nav_mobile.html
```diff
@@ -1,15 +1,13 @@
-<!-- 
-  Copyright (c) 2018 Eclipse Foundation, Inc.
+<!--
+ Copyright (c) 2025 Eclipse Foundation AISBL
```
ericpoirier: Same here.

gnugomez: those are new files, should I keep the year from the original file (they have plenty of new markup)? then I will need to keep the Inc. too, right?

ericpoirier:
> any suggestion on how to fix this @ericpoirier ?

@gnugomez related to the menu alignement, have you tried moving the logo instead of the menu. Might be easier to move/resize then the whole menu?

oliviergoulet5:
> top

image

image

image


It looks fine for me when I try to align the logo and text. I'm just adding margin-bottom to the logo and changing the `margin-bottom` for `.navbar-wrapper:has(+ header .jumbotron)`

gnugomez:
@oliviergoulet5 @ericpoirier This adds an odd white space just under the logo; it might feel more aligned, but we’re not following the design provided.

autumnfound:
> @oliviergoulet5 @ericpoirier This adds an odd white space just under the logo; it might feel more aligned, but we’re not following the design provided.

Rather than block development on this, I suggest we merge this in as-is and approach Elspeth to see if she's open to updating the design to account for this change. @gnugomez make sure you make a note to address this concern, and otherwise we can defer this so you can continue work while those discussions happen!

oliviergoulet5 (approved):

PR #2297, opened 2025-11-21
feat: redesigning footer

gnugomez:


less/components/footer.less
```diff
@@ -0,0 +1,92 @@
+footer#solstice-footer {
```
autumnfound, line 11: Missing copyright header

layouts/partials/footer_links/suffix.html
```diff
@@ -0,0 +1,21 @@
+<section class="col-sm-24 col-md-9 padding-top-20 social">
```
autumnfound, line 11: Missing copyright header

layouts/partials/footer_links/footer-newsletter.html
```diff
@@ -0,0 +1,37 @@
+{{ $footer_newsletter_class := .Params.footer_section_newsletter_class | default .Site.Params.footer_section_newsletter_class | default "footer-section col-sm-8" }}
```
autumnfound, line 11: Missing copyright header

less/components/footer.less
```diff
@@ -0,0 +1,92 @@
...
+      }
+    }
+  }
+
+  .footer-end-newsletter {
+    padding: 3rem;
+    border: 1px solid @brand-primary-orange;
+    border-radius: 20px;
+
+    @media (max-width: @screen-sm-max) {
```
autumnfound: Is there a reason we are using desktop-first CSS here? We tend to use mobile-first wherever possible.

layouts/partials/footer.html
```diff
@@ -27,6 +29,4 @@
       {{ partial "footer_links/footer-copyright.html" . }}
```
autumnfound, line 29: The copyright section of the footer seems to be missing the design call out, was that removed from the final site?

oliviergoulet5:
+1 (aside from Martin's feedback!)

less/components/footer.less
gnugomez: not really I guess it came out like this, since the only design we have is the desktop one; I started from there

layouts/partials/footer.html
gnugomez: I'll add it ◡̈

autumnfound (approved):
LGTM!

Issue #2240, opened 2025-09-30
Implement the new Jakarta EE website homepage 


shabnammayel:
As discussed on the call today, please find the Figma files here: 

https://www.figma.com/design/2KewKkyuXUqa8DzBoW9rBN/SHARED_Jakarta-EE-New-Branding-AUG-19-2025?node-id=2219-1394&p=f&t=DSQtAYm8wMXCRKsv-0

I am tagging @excentrickristy here in case you need any additional assets for the designs for implementation

Thanks

autumnfound:
Hi all, just some follow up questions that @gnugomez and I had after spending some time looking through the Figma design.

1. For the nav menu, what is the intention with mobile view of the menu? Do we want to use the normal functionality with collapsible menu items, or is there some other design/functionality intended for the menu at mobile and tablet views?
1. Additional question for the nav menu. In the expanded views of the nav menus, there is a white banner/background under the header, but in other views that banner is not there. Is that an older design element that was left after reworks of the main page, or is there some intended functionality to make that banner appear in some situation?
1. For the GitHub contributor section, we just want to confirm that the profile bubbles are static, or if not what is the intention behind that, as it's hard to convey in a static design. As an additional followup, what is the plan for smaller than desktop views? The floating profile pictures would likely be very cluttered on anything less than desktop, so do we want to keep that as a sort of flair for normal desktop view and just have a standard banner section in those breakpoints?
1. For the testimonials section under Join Us, what is the intended display here? Is the intent for it to be edge to edge? Do we want manual scrolling, a slow slide to loop the testimonials, or something else?
1. For the sticky footer, do we intend to force the user to have the expanded view in the corner at all times after initial scroll? That seems like it might block some content and might impede the user experience if we don't have way of collapsing it back down to the initial view, such as a little collapse arrow.
1. Regarding the cookie consent banner, right now it remains the solid grey banner with orange buttons which somewhat clash with the new design, and there doesn't appear to be an alternative design provided in the Figma files. Jordi had a good suggestion of using the new cookie consent banner that is being created for the new EF look and feel we are currently implementing. I don't know where he pulled the design screenshots from, though. Jordi, could you post a screenshot of the new version so we can get see if that would work with the resdesign?

Also, Jordi may have more questions WRT the design elements than I do, so he may follow up with more questions in the next day.

shabnammayel:
[...]

PR #2302, opened 2025-11-26
feat: redesigning home

gnugomez:
Related to #2240


oliviergoulet5:
**issue**: I'm spotting a z-index problem here. The button and some text content from one of the sections are appearing over the navbar: 

image

oliviergoulet5:
**praise**: The animations are really nice! And the page looks great! Good job @gnugomez

oliviergoulet5:
**issue**: There are accessibility issues with this section & buttons at smaller screen widths:

image

oliviergoulet5:
**suggestion(non-blocking)**: The horizontal scroll section with testimonials should probably auto-scroll every 5-10 seconds. Otherwise, there is no great indicator to suggest it's a scrollable section.

Either it auto-animates, or we add actual physical button triggers or other visual clues (e.g. arrows or fade-out gradients) to indicate it's scrollable.

But you should check-in with the design team.

gnugomez:
@oliviergoulet5 any suggestion on how to fix the contrast problem with the buttons?

gnugomez:
@oliviergoulet5 the design team shared this link in regards the horizontal scroll-able sections https://webflow.com/made-in-webflow/website/simple-horizontal-cards

They apparently want the user to lead the movement, so I won't make them auto-scroll, although I added some controls to the two scroll-able carousels, just to make them more accessible, I'm adding the respective aria attributes too.

autumnfound:
Praise: The contributor bubbles look really good, and overall the paging controls for the carousels work fairly well without changing the design.

autumnfound:
comment: RE the controls and auto-scroll, I think it's worth having some discussions w/ marketing about those. I don't think it's worth blocking on since we have a serviceable solution that doesn't break spec.

For the contrast on buttons, I think it's the same thing. This matches spec and we definitely need to have the conversation, but that can be part of the feedback rounds.

oliviergoulet5:
**comment**: RE the contrast issues with buttons

>any suggestion on how to fix the contrast problem with the buttons?

I see the following options:

1. Tweak background position at specific breakpoints
2. Apply a `mix-blend-mode` (e.g. `color-dodge`) - although this affects the visuals significantly
3. Request a more contrast-friendly design

My concern with 1 and 2 would be that contrast issues will forever be on our mind when using outline buttons. It's a bit of an annoyance. But TBH it isn't uncommon to have outline buttons, so we shouldn't need to resort to 3.

content/_index.zh.md
oliviergoulet5, line 1: **question**: Do you have a plan to translate _index.html? 

I've always thought about adding an `i18n` shortcode which would be a shortcode wrapper for Hugo's `i18n` function.

However, that would require a way for us to generate the multilanguage page without creating a file for it. Perhaps there is a Hugo param for outputting a page in multiple languages.

oliviergoulet5 (approved):

content/_index.zh.md
oliviergoulet5: Otherwise, perhaps the page markup should be in `layouts/`

IIRC, you can create a layout file for `Home` pages. Or just keep it simple and make it a shortcode page.

gnugomez: true, I was thinking about following the same approach as we were following before of duplicating the content of the page but with translations, but quite sincerely that will lead to inconsistencies as it did before, so I'll follow the shortcode approach you mention ◡̈.

for the menu I'm trying to follow a similar approach since Hugo apparently allows you to define translation keys for menu items

PR #2308, opened 2025-12-04
feat: adding secondary menu + menu design updates

gnugomez:


layouts/partials/menu_secondary.html
```diff
@@ -0,0 +1,13 @@
+{{ $class := "" }}
```
autumnfound, line 11: Missing copyright header

oliviergoulet5:
**issue**: The dropdown for language and the cookie link have no focus outline

layouts/partials/menu_secondary.html
gnugomez: 😅

autumnfound (approved):

PR #2311, opened 2025-12-04
2025 redesign - Jumbotron, typography, megamenu fixes

autumnfound:


config/_default/menu.ja.yaml
```diff
@@ -61,91 +55,6 @@ main:
     pageRef: "/about/"
     weight: 6
 
-  # ========================================
```
gnugomez, line 64: **question:** I'm a bit confused, why we remove this menu item?

I guess it's somewhat related with the fact that some pages aren't developed yet, but here there was the learn page links and some others

autumnfound: Sorry, yes this was a request from Shabnam when we had a call yesterday!

oliviergoulet5 (approved):

autumnfound:
Merging w/o squash so that we can easily revert the megamenu update when we want to go live with the developer portal

oliviergoulet5:
**question(non-blocking)**: I see there’s a `text-primary-orange` utility. Creating two “primary” colours seems like it could get messy. I know swapping primary and secondary might require a significant refactor. But is duplicating the primary color really the best approach here? Why not use secondary? This feels like it would introduce technical debt. I’m also not a fan of tying colour names to their actual hues, since it makes future branding changes harder.

I set this as non-blocking because it's probably out of scope of the patch.

autumnfound:
> Creating two “primary” colours seems like it could get messy

I agree, but it comes down to how the design + branding for the site was created. In this case, the design indicates 5 different primary colours and there isn't a simple pattern like we have for our internal designs, and they get interchanged in different scenarios. To support these cases, we have multiple primaries. It isn't ideal, but it makes matching the design spec much cleaner.

PR #2317, opened 2025-12-09
feat: adding platform diagram

gnugomez:
This section pops up as soon as you press the view specification button from the Jakarta EE 11 banner 

image


autumnfound:
General note, for files that are AI assisted in development, we should be putting a comment into the file to indicate that the file was at least paritally generated by an AI along with which AI https://www.eclipse.org/projects/guidelines/genai/

autumnfound (approved):

PR #2318, opened 2025-12-09
fix: Update homepage events CTA, fix header size + spacing

autumnfound:


autumnfound:
When fixing the subpage headers in a previous patch, it made the homepage headers too small as well. To compensate, I added a new class that should allow us to easily apply the featured heading style to get the font weight, size, and letter spacing as defined in the design files.

layouts/shortcodes/pages/home/events.html
```diff
@@ -2,25 +2,20 @@
 <section id="events" class="row padding-y-60" aria-labelledby="events-heading">
   <div class="container">
     <div class="row">
-      <div class="col-xs-24">
```
gnugomez, line 5: **question:** was that a request from Shabnam?

layouts/shortcodes/pages/home/events.html
```diff
@@ -2,25 +2,20 @@
...
+      <div class="col-xs-24 padding-30">
+      <div class="grid md:grid-cols-2 gap-y-line-4 md:gap-x-line-4 align-items-center">
+        <div class="text-center">
+          <h2 id="events-heading" class="text-primary featured-heading margin-top-0">
+            {{ i18n "home-events-heading" }}
+          </h2>
+          <a href="/community/events/" class="btn btn-secondary" aria-label="{{ i18n "home-news-read-all-news-aria" }}" target="_blank" rel="noopener noreferrer">{{ i18n "home-events-view-all-events" }}</a>
+        </div>
+        <p class="text-primary">
+          <strong>Join developers, contributors, and Jakarta EE enthusiasts</strong> from around the globe as we gather — online and in person — for a series of upcoming events, meet-ups, and livestreams.
```
gnugomez: **issue:** this should be a translation key I would suggest

layouts/shortcodes/pages/home/jakarta-ee-11-release.html
gnugomez, line 1: **note:** I've updated this file in another PR, this will most likely produce conflicts

less/components/typography.less
```diff
@@ -102,6 +102,12 @@ h6,
   .text-medium;
 }
 
+.featured-heading {
```
gnugomez, line 105: **nitpick:** shouldn't this class also define color?

I see that most usages of this are using the text-primary utility

layouts/shortcodes/pages/home/events.html
autumnfound: It was, yes. It was added to the latest design mockup

less/components/typography.less
autumnfound: Most usages yes, but I think leaving it out gives us the most flexibility as it allows us to apply/leave colour as is. While most do use that colouring, this was added after as an enhancer

gnugomez (approved):

PR #2324, opened 2025-12-11
feat: adding featured story

gnugomez:


ericpoirier (approved):

oliviergoulet5:
**issue**: All looks good except for keyboard navigation. I have a hard time seeing the focus. I think it's actually focusing the button before the featured story even opens. 

IMO, upon focus, the featured story should automatically expand.

gnugomez:
> **issue**: All looks good except for keyboard navigation. I have a hard time seeing the focus. I think it's actually focusing the button before the featured story even opens.
> 
> IMO, upon focus, the featured story should automatically expand.

Yeah, I think it's a fair point, I'll make it open just by hover or focus within do you think we should add a tabindex to the whole thing @oliviergoulet5 ?

oliviergoulet5:
I would try to follow the success criterion of this article: https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html

I think this page best represents what you're trying to do here.

oliviergoulet5:
I think adding a tabindex for the whole thing would work honestly

oliviergoulet5 (approved):

PR #2338, opened 2025-12-12
feat: opening external links in new tab

gnugomez:


layouts/partials/menu_link.html
```diff
@@ -0,0 +1,20 @@
+{{/*
```
autumnfound, line 11: **nitpick**: Missing copyright header

autumnfound (approved):
Otherwise LGTM

PR #2353, opened 2026-01-06
fix: Update fonts CORS headers in Netlify to cover TTF + WOFF files

autumnfound:


autumnfound:
This is to address an issue where trailing content on the file name request would cause font lookups to fail externally, namely on the start.jakarta.ee site. Additionally, we also need to support TTF files as well as WOFF files.

ericpoirier (approved):

PR #2356, opened 2026-01-09
chore: versioning styles to avoid issues from consumers

gnugomez:
In this PR I'm basically checking out the js and less folder from `src` to `js/v1` and `less/v1` and moving the redesing assets to `js/v2` and `less/v2`, the existing outputs represent the "Legacy" design, the `v2` ones the redesign.


gnugomez:
I've been testing this approach with the devtools on `start.jakarta.ee` and it's working fine.

ericpoirier (approved):

autumnfound (approved):

PR #2360, opened 2026-01-12
fix(home): h1 has unnecessary whitespace and requires screen reader text

### LINKED ISSUE #2359: Animated text in homepage jumbotron has right padding
> Congrats and thank you @gnugomez and @autumnfound for the new website! 
> 
> I am looking at the banner, is there two spaces between the rotating words and enterprise? if yes,can you please remove one?
> 
> Image 

 _Originally posted by @shabnammayel in [#2240](https://github.com/jakartaee/jakarta.ee/issues/2240#issuecomment-3739696311)_

With some initial investigations, this seems to be caused by an interaction between the `letter-spacing` and the animation styling. Removing the letter spacing on the section removes the right padding after the content, but means that the text doesn't have consistent styling, which just creates a new issue.

oliviergoulet5:
Resolves #2359


gnugomez:
Thanks for the patch @oliviergoulet5, I'm pushing some new code to make this rotating text more reliable.

I rewrote the whole thing since it was a bit brittle only using css, now I'll be using js for widths and stuff.

content/_index.ja.md
```diff
@@ -4,7 +4,14 @@ seo_title:
   "Jakarta® EE | クラウドネイティブエンタープライズJava |Java EE |Eclipse
   Foundation"
 headline: |
-  Jakarta EE powers the future of <br><span class="text-primary-orange rotating-text"><span>cloud native</span><span>vendor neutral</span><span>scalable</span></span> enterprise Java
+  <span aria-hidden="true">Jakarta EE powers the future of <br><span class="text-primary-orange rotating-text-container">
+  <span class="rotating-text-items">
```
ericpoirier, line 8: **nitpick:** I think we should use something like aria-hidden="true" here since on line 14 we have a sr-only span.

oliviergoulet5:
**issue**: mobile has some serious text overlap 

image

content/_index.ja.md
gnugomez: where do you mean? we already have area-hidden when wrapping the text

gnugomez:
> **issue**: mobile has some serious text overlap
> 
> image

@oliviergoulet5 Have you tried reloading the page? the word size is calculated on load, so if you have resized the window it might have produced some bugs, I can't reproduce you're screenshot.

oliviergoulet5 (commented):
LGTM

autumnfound (approved):

PR #2374, opened 2026-01-20
feat: adding join-us and developer portal pages

### LINKED ISSUE #2326: Create Join us and Developer Portal pages
As part of the 2025 redesign, we are introducing new "Join Us" and "Developer Portal" pages. These pages design mockups will be provided through Figma by @shabnammayel. As part of these changes, we will want to reintroduce the megamenu section for the developer portal and the jumbotron CTA link.

The targeted release date of these pages is by end of January 2026, to be released as completed and approved.

- [ ] Create Join us page
- [ ] Create Developer Portal page

gnugomez:
Resolves #2326


autumnfound:
IT looks like the jumbotron default image is way too high, leading to only the lower half of the jumbo image background being visible. Otherwise LGTM

gnugomez:
> IT looks like the jumbotron default image is way too high, leading to only the lower half of the jumbo image background being visible. Otherwise LGTM

that was the idea, or at least what I can take from the design, the white gradient is part of the bottom of the image, so I'm aligning it like that, if the jumbo grows it will reveal a little more image from the top.

autumnfound:
> that was the idea, or at least what I can take from the design, the white gradient is part of the bottom of the image, so I'm aligning it like that, if the jumbo grows it will reveal a little more image from the top.

That's honestly the opposite of what I see in the design:

image

Compared to this:

image


The gradient starts around the bottom of the jumbotron title in the design, and not below the megamenu.

autumnfound:
Example fix with CSS `background: linear-gradient(to bottom, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0) 59%, rgb(255, 255, 255) 100%), url(../images/jumbotron-bg-default.webp?11e7e245a4f2f809a55a025f641f3da2)` + removing the position bottom:
 
image

gnugomez:
> Example fix with CSS `background: linear-gradient(to bottom, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0) 59%, rgb(255, 255, 255) 100%), url(../images/jumbotron-bg-default.webp?11e7e245a4f2f809a55a025f641f3da2)` + removing the position bottom:
> image

But that's a different case, because the one you shared from the design has clearly a bigger jumbotron, we could set that size by default if you think is that much of a problem, but shifting the image down reduces the space for the white gradient and makes the transition too obvious, it's quite harsh I would say.

gnugomez:
image

something like this.

another option could be to extract the image without gradient, keep it centered and add the white gradient attached to the bottom with css, I'm not sure about that one tho, there's not just one gradient, there's also some other round gradients floating in random positions, I feel like it's all part of a single composition, and shifting it around will make it look strange.

gnugomez:
- https://deploy-preview-2374--jakartaee.netlify.app/developer-portal/
- https://deploy-preview-2374--jakartaee.netlify.app/join-us/

autumnfound:
> But that's a different case, because the one you shared from the design has clearly a bigger jumbotron, we could set that size by default if you think is that much of a problem, but shifting the image down reduces the space for the white gradient and makes the transition too obvious, it's quite harsh I would say.

Sorry, I had missed these comments! I agree my solution isn't a perfect one, but I think it highlights our real problem: We need to figure out a universal treatment of jumbotron banner image/background for non-home pages. We have 3 different look and feels, with one of the newer pages having another look and feel for us to provide. I think it might be worth having a discussion with the design folks to figure this out!

## Developer portal

The glyphs in the tiles for the swapping sectional portion of the page seems to be missing. The screen shots I see from previous previews had those in, I'm guessing that was a conversation with Shabnam to remove them?

Not sure I love how this looks at tablet breakpoint, I'm thinking centering and making the buttons bigger might help a little. As well, it might be worth having a slightly higher quality image if that's possible just so it doesn't look super blown out. Ignore the random menu bar, just a quirk of how I took the screenshot.

image

## Join us

The sm breakpoint is instead breaking at the md breakpoint, which is leading to blown out images for the first section tiles. The shadow on those tiles is also much weaker than the design.

The testimonials also diverge from the design, but I personally like the consistency more in this case. We also have a bunch of broken logos in the testimonial section we should fix:
image

js/v2/scroll-snap-carousel.js
```diff
@@ -8,126 +8,195 @@
...
+  container.addEventListener("mouseenter", () => (isHovering = true));
+  container.addEventListener("mouseleave", () => (isHovering = false));
+
+  // Start auto-rotate if enabled
+  if (options.autoRotate) {
+    autoRotateTimer = setInterval(() => {
+      if (!isHovering) {
+        move(1);
+      }
+    }, options.autoRotateInterval || 3000);
```
autumnfound: This is _way_ too fast. I'd say we should probably start at least at 6-7 seconds and maybe go up from there. At this pace it's hard to read, and isn't mobile accessible as there is no hover state on mobile

gnugomez: The idea is that if you hover over the moving cards they will stop rotating, so that you can actually read it, but I could slow it down a bit, but maybe not as much as 6-7 seconds maybe 4 or 5

gnugomez:
@autumnfound  the glyphs not showing up was not a deliberate decision; it was an oversight on my part. I adapted the pages to be multilingual just before merging, which resulted in this issue

gnugomez:
@autumnfound as a solution for the events cards in tablet, I added two cols so that the look a bit more balanced

image

gnugomez:
@autumnfound, I know the testimonials diverge from the design but the actual design is not too far away, and I'd better have a bit more consistency here. if they complain I can implement the different design. 

As per the logos I've pushed an update to avoid adding the "'s logo" since in some cases they're just self employed folks, thus we can keep the alt text for those cases. I also made that text fit a bit better, otherwise it was a bit hard to read

js/v2/scroll-snap-carousel.js
autumnfound: The important part there is that on mobile there is no way to hover, so this impacts accessibility. Also, it isn't intuitive that hover stops the scroll, so we shouldn't rely on that.

gnugomez: indeed it isn't a made up thing, this tends to be the standard behaviour when something rotates. in mobile should be the same, hovering your finger over should prevent rotation. this is also what instagram does for their "stories"

autumnfound:
image

Otherwise I'm not seeing any glaring issues

autumnfound (approved):
I don't see anything wrong here, but it might still be good to get Olivier to look at it

oliviergoulet5 (approved):

PR #2398, opened 2026-02-25
feat: adding release 12 page

### LINKED ISSUE #2330: Create new Jakarta EE 12 release page
As part of the normal release cadence, we are creating a new release page for Jakarta EE 12. The design for this page is to be delivered through Figma by @shabnammayel by end of January 2026.

The target due date for this page is by end of February 2026, to be released as completed and approved by stakeholders.

gnugomez:
Resolves #2330

### DIFF
--- a/content/release/12/index.html
+++ b/content/release/12/index.html
@@ -1,6 +1,7 @@
 ---
 title: "Jakarta EE 12"
 seo_title: "Jakarta EE 12"
+date: 2026-02-26T00:00:00 
 headline: |
   <div aria-hidden="true" class="release-title">
     <div class="title">
@@ -24,6 +25,8 @@ links:
     ],
   ]
 links_position: "before_custom_jumbotron"
+description: Jakarta EE 12 is under development. Track proposals, milestones, and specification progress here.
+under_development: true
 hide_page_title: true
 hide_sidebar: true
 hide_breadcrumb: true
--- a/content/release/12/index.html
+++ b/content/release/12/index.html
@@ -29,7 +29,7 @@ hide_sidebar: true
 hide_breadcrumb: true
 ---
 
-<section class="row text-primary">
+<section class="row text-primary padding-y-60">
   <div class="col-md-24 margin-bottom-60">
     <div class="max-w-50 margin-x-auto text-center">
       <p>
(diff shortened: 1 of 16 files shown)


autumnfound:
I'd definitely recommend going to stacked rather than side by side. The side by side is cramping the text in the spec diagram and making it feel way more busy. Additionally, we will likely want to make the text a floating block that is half width and centered to make it feel less small.

Additionally, this page uses an alternate background in the design than the standard image that is being used.

gnugomez:
> I'd definitely recommend going to stacked rather than side by side. The side by side is cramping the text in the spec diagram and making it feel way more busy. Additionally, we will likely want to make the text a floating block that is half width and centered to make it feel less small.
> 
> Additionally, this page uses an alternate background in the design than the standard image that is being used.

As per the layout, I’ve removed some padding from the diagram so now it looks completely fine. 

I tried what you suggested whilst building the page, but it looked quite bad, since all in the page is already centred.

As per the background, I wouldn’t pay too much attention to that, since they’ve changed the design quite a lot and the background from the design is the initial one they had; they now just use it for the homepage, so that’s why I used the one we use everywhere else here.

autumnfound:
> I tried what you suggested whilst building the page, but it looked quite bad, since all in the page is already centred.

Agreed, it looks fairly bad if you just center the text in a full width container. What I'm saying is we should add a max-width to the text, center it, and stack it on top like in the screenshots below. This makes the text feel important and not anemic like the full-width content feels.

This gives both pieces of information some breathing room in our container, giving the diagram some extra space so we don't have multi-line spec items at any width, and making the preamble have some level of importance.

image
image

ericpoirier (approved):

content/release/12/index.html
```diff
@@ -0,0 +1,50 @@
...
+  </div>
+  <span class="sr-only">Jakarta EE 12 - under development</span>
+subtitle: Community-driven innovation<br> for <strong>robust and flexible</strong> enterprise Java
+jumbotron_btn_class: btn btn-secondary
+jumbotron_class: text-center
+custom_jumbotron_class: col-sm-18 col-sm-offset-3 margin-top-60
+links:
+  [
+    [
+      href: "/platform/12/",
```
autumnfound: The links here are broken: https://jakarta.ee/specifications/platform/12/

autumnfound (approved):
LGTM

PR #2423, opened 2026-03-26
feat: updating about us page

### LINKED ISSUE #2328: Update existing "About" page
As part of the 2025 redesign, we are updating the existing "About" page to proved more information about governance and technology related to the Jakarta EE working group. The design for this page is to be delivered through Figma by @shabnammayel in January 2026.

The target due date for this page is by end of March 2026, to be released as completed and approved by stakeholders.

gnugomez:
Resolves #2328


layouts/partials/developer-portal/icon-grid-section.html
autumnfound, line 1: **nitpick(non-blocking)**: Is there a reason we are labeling this as a `developer-portal` partial? This seems to be used in multiple sections now, and might be worth renaming to be clear that this is a shared component.

layouts/shortcodes/pages/about/why-it-matters.html
```diff
@@ -0,0 +1,40 @@
...
+ This program and the accompanying materials are made
+ available under the terms of the Eclipse Public License 2.0
+ which is available at https://www.eclipse.org/legal/epl-2.0/
+
+ SPDX-License-Identifier: EPL-2.0
+-->
+
+<section class="row position-relative padding-y-large">
+  <div class="container">
+    {{ partial "developer-portal/icon-grid-section" (dict
```
autumnfound: **suggestion (non-blocking)**: We should consider making these data files instead as tech debt. This is a fairly ugly way to print this out, this just looks like yaml written the hard way.

autumnfound (changes requested):
Additionally, the updated page is totally inaccessible due to a redirect: https://github.com/jakartaee/jakarta.ee/blob/src/static/_redirects#L19

layouts/shortcodes/pages/about/why-it-matters.html
gnugomez: yeah, I agree. I was just trying to make content blocks more reusable so that if the markup changes, it would be easier to refactor, but the Go-tmpl syntax is what it is...

I will just use the actual markup; using data files is a bit of an overkill for this and it makes it difficult to trace.

less/v2/components/utilities.less
```diff
@@ -360,3 +360,39 @@
...
+  --fade-in-delay: 1s;
+}
+
+.delay-3 {
+  --fade-in-delay: 1.5s;
+}
+
+.delay-4 {
+  --fade-in-delay: 2s;
+}
```
oliviergoulet5: **question**: Maybe I'm just not finding it, but are any of these `fade-in` classes and animations used in your patch? I see you're using animations from EFSA, so maybe these need to be cleaned up?

gnugomez: oh right, forgot to clean that.

let me do so

gnugomez: updated

oliviergoulet5:
**praise**: Aside from my comment above, +1. Looks good!

PR #2444, opened 2026-04-30
chore(i18n): adding support for vscode i18n ally extension

gnugomez:
This extension eases quite a lot the translation efforts needed, it provides a really good visualization of the translations, and it integrates with deepl for automatizing translations.


config/_default/menu.ja.yaml
ericpoirier, line 1: **question:** you don't need those default files anymore?

gnugomez: now the menu labels are just i18n keys, this is an approach we're taking to prevent other languages of the menu from being stale, now there's just one menu definition, and the labels can be translated

autumnfound (approved):
LGTM

